PR #91, opened 2014-10-21
ASM-2540 Don't remove RAID if it doesn't exist

gavin-scott:


gavin-scott:
Hmm, I'm not sure why this is here. Closing since it hasn't been reported as an 8.0 problem.

PR #105, opened 2015-01-15
Changed importsystemconfiguration workflow to help reliability/stability

asaenz:
There will now be 4 files from the exports/imports.  
  * <servicetag>_original.xml which contains the target server's configuration before any changes are made
  * <servicetag>_preset.xml which contains basic attributes that we send in first to "preset" the idrac to be in a state that's easy for us to use.  This includes setting up partitioning, setting up boot protocol, enabling raid, enabling SD card.
  * <servicetag>_base.xml which contains the configuration that will be used as the reference for the config we send to the server
  * <servicetag>.xml which contains the system configuration we import to the target server.

The main part is the _preset.xml, which should allow us to enable/disable what needs to be set up on the idrac in order for the import to work as smoothly as possible.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -8,7 +8,7 @@ include REXML
 
 class Puppet::Provider::Importtemplatexml <  Puppet::Provider
 
-  def initialize (ip,username,password,resource)
+  def initialize (ip,username,password,resource, exported_postfix='base')
     @ip = ip
     @username = username
     @password = password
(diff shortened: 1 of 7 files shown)


gavin-scott:
Tests failed. Good luck!

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -18,11 +18,58 @@ def initialize (ip,username,password,resource)
...
   end
 
   def importtemplatexml
     munge_config_xml
-    response=executeimportcmd
+    executeimportcmd
+  end
+
+  #Bugs with the ordering of attributes/components in the import causes issues. 
+  #This function will go send a very basic xml that will set attributes such as IntegratedRaid, InternalSdCard, NicPartitioning, etc so they are ready and valid in our big import.
```
gavin-scott: Maybe put a TODO in here that we should check whether we can skip this step entirely, e.g. if the server is already set up properly.

asaenz:
@gavin-scott Tests are passing.

PR #106, opened 2015-01-27
ASM-2857 Reset virtual mac addresses during deployment

gavin-scott:
Should help brown-field cases where the server has virtual mac
addresses set. Specifically this fixes a case we saw where they
had a server with duplicate virtual mac addresses.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -517,6 +517,14 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
                 end
               end
 
+              # Reset virtual mac addresses by default
+              if changes['NicMode'] == 'Enabled'
+                changes['VirtMacAddr'] = '00:00:00:00:00:00'
+              end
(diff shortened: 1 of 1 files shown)


gavin-scott:
@sushilrai please review.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -517,6 +517,14 @@ def process_nics
                 end
               end
 
+              # Reset virtual mac addresses by default
+              if changes['NicMode'] == 'Enabled'
+                changes['VirtMacAddr'] = '00:00:00:00:00:00'
```
sushilrai: We need to set the virtual MAC address same as physical MAC address.

gavin-scott: What we've seen is that setting it to 00:00:00:00:00:00 resets it to the physical mac address. @asaenz already added teardown code that relies on that:

https://github.com/dell-asm/dell-idrac/blob/e635c9ab6c8966f6475d8108d2bb265a28490cd9/lib/puppet/provider/importtemplatexml.rb#L298

PR #107, opened 2015-02-03
Support for user defined raid configuration

asaenz:
MUST BE MERGED WITH: https://github.com/dell-asm/asm/pull/211 and https://github.com/dell-asm/asm-deployer/pull/742


gavin-scott:
CI tests failed.

lib/puppet/idrac/util.rb
```diff
@@ -1,19 +1,43 @@
...
+require 'pty'
+
+module Puppet
+  module Idrac
+    module Util
+      def self.get_transport
+        require 'asm/device_management'
+        @transport ||= begin
+          t = ASM::DeviceManagement.parse_device_config(Puppet[:certname])
+          t[:password] = URI.decode(t[:password])
```
gavin-scott: This shouldn't be necessary, parse_device_config handles the password:

https://github.com/dell-asm/asm-deployer/blob/fb6ddc64e1e6681b060bef5ddde48a5310f6df06/lib/asm/device_management.rb#L201

Were you running into a scenario where you needed to do this?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -141,7 +143,7 @@ def munge_config_xml
 
     if(!raid_in_sync?(xml_base))
       #Need xml_base to check raid_config_changes, so need to do here instead of in get_config_changes
```
gavin-scott: Is this comment out of date now that xml_base has been removed as an argument for get_raid_config_changes?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -288,6 +290,7 @@ def munge_virt_mac_addr(nc, changes)
   # it in the list of components.
   #
   # Typical values will be RAID.Integrated.1-1 or RAID.Slot.6-1
+  # TODO:  This will need to change to return a list of controllers in the future for multiple controller support.
   def raid_controller
```
gavin-scott: Isn't the controller passed in with the raid_configuration data? Why do we still need to find a raid controller to use?

lib/puppet/idrac/util.rb
asaenz: @gavin-scott This is actually not my code.  I think I changed the line endings so it looks like I made this change, but that was already there.  I can't speak on why it's necessary.

lib/puppet/provider/importtemplatexml.rb
asaenz: @gavin-scott The comment is a little misleading now, but it still mostly applies since I still have to do a raid_in_sync? check with the xml_base beforehand.  I'll update it slightly.  There could definitely be some refactoring around xml_base I think (since xml_base is now a method that gets the xml_base, and we should theretically always have an xml_base to compare against).  I'm not sure that's appropriate for this PR/time frame, though.

lib/puppet/idrac/util.rb
gavin-scott: It's not, can you remove it?  Thanks.

lib/puppet/provider/importtemplatexml.rb
gavin-scott: OK, that's fine.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -302,74 +305,140 @@ def raid_controller
...
       Puppet.debug("RAID_ACTION: #{@resource[:raid_action]}")
-      changes['whole']['RAID.Integrated.1-1'] =
-          {
-              'RAIDresetConfig' => "True",
-              'Disk.Virtual.0:RAID.Integrated.1-1' =>
-                  {
-                      'RAIDaction'=>'Delete',
-                  }
-          }
+      changes['whole']['RAID.Integrated.1-1'] = { 'RAIDresetConfig' => "True" }
```
gavin-scott: This should probably be all raid controllers that are referenced in the raid_configuration

spec/unit/puppet/provider/importtemplatexml_spec.rb
```diff
@@ -135,7 +193,22 @@
       xml.xpath("//Component[@FQDD='RemoveMe']").size.should == 0
       xml.xpath("//Component[@FQDD='BIOS.Setup.1-1']/Attribute").first.content.should == "Disabled"
       xml.xpath("//Component[@FQDD='LifecycleController.Embedded.1']/Attribute").size.should_not == 0
-		end
+    end
+
+    it "should get changes based on raid configuration hash" do
```
gavin-scott: Yay test!

asaenz:
@gavin-scott Tests are passing.  Might want to give a quick review again, since I made a couple changes based on your comments earlier.

PR #109, opened 2015-02-09
Ensure RAID is enabled on teardown.

asaenz:
MUST BE MERGED WITH: https://github.com/dell-asm/asm/pull/218 and https://github.com/dell-asm/asm-deployer/pull/766

### DIFF
--- a/lib/puppet/provider/idrac.rb
+++ b/lib/puppet/provider/idrac.rb
@@ -22,7 +22,7 @@ class Puppet::Provider::Idrac <  Puppet::Provider
       reset
       exporttemplate
     end
-    synced = !resource[:force_reboot] && config_in_sync?
+    synced = resource[:ensure] != :teardown && !resource[:force_reboot] && config_in_sync?
     Puppet.info("Server is already configured.  Skipping import...") if synced
     synced
   end
@@ -132,7 +132,7 @@ class Puppet::Provider::Idrac <  Puppet::Provider
         elsif(existing_val != value)
           if(key == "BiosBootSeq")
             compare = value.delete(' ').split(',').zip(existing_val.delete(' ').split(',')).select{|new_val, exist_val| new_val != exist_val}
-            if(compare.size != 0 && @resource[:raid_action] != :delete)
+            if(compare.size != 0 && @resource[:ensure] != :teardown)
               Puppet.debug("Value of BiosBootSeq does not match up. Existing Seq: #{existing_val}, trying to set to  #{value}")
               in_sync = false
               break
--- a/lib/puppet/provider/importsystemconfiguration/default.rb
+++ b/lib/puppet/provider/importsystemconfiguration/default.rb
@@ -13,8 +13,15 @@ Puppet::Type.type(:importsystemconfiguration).provide(
 
   def create
     instance_id = setup_idrac
-    wait_for_import(instance_id)
-    #Need to export again so we have a template with updated values from the setup step before.
+    wait_for_import(instance_id) if instance_id
+    import_config
+  end
+
+  def teardown
+    import_config
+  end
+
+  def import_config
     exporttemplate('base')
     instance_id = importtemplate
     wait_for_import(instance_id)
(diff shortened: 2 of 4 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -115,7 +115,10 @@ def default_changes
         }
     changes['whole'] = { 'LifecycleController.Embedded.1' => { 'LCAttributes.1#CollectSystemInventoryOnRestart' => 'Enabled' } }
     # target_boot_device settings
-    if @resource[:target_boot_device] == "HD"
+    #Always want to turn on IntegratedRaid with a delete, so ASM can continue to inventory RAID info later.
+    if @resource[:raid_action] == :delete
+      changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'IntegratedRaid' => 'Enabled'})
```
gavin-scott: This seems like a weird place to put this ASM-specific logic to me. Could we pass this in as a generic BIOS option to the asm::idrac resource from the asm-deployer teardown code, i.e. here instead:

https://github.com/dell-asm/asm-deployer/blob/89c772b0a8b830f3ebb710f996b985cc3fb61506/lib/asm/service_deployment.rb#L4385

lib/puppet/provider/importsystemconfiguration/default.rb
```diff
@@ -12,9 +12,12 @@
   desc "Dell idrac provider for import system configuration."
 
   def create
-    instance_id = setup_idrac
-    wait_for_import(instance_id)
-    #Need to export again so we have a template with updated values from the setup step before.
+    unless resource[:raid_action] == :delete
```
gavin-scott: I don't really get the conditional here. Is this a poor man's version of actually checking whether we need to enable/disable things in the "pre-import" step?

gavin-scott:
Ping me if you want to discuss this further.

lib/puppet/provider/importsystemconfiguration/default.rb
asaenz: @gavin-scott This is just because this code gets a little weird at this point when deleting.  Normally, we want to make sure certain things are set up before we do any import (such as turning on integrated raid, before we try to configure raid).  However, in the case of delete, we only need to run once because we're just cleaning up (which only requires a single import).  
This setup step is not necessary while deleting, so it can be skipped.

asaenz: The reason it gets weird is because there's almost no difference in what we do between deleting and creating other than we pass in "raid_action=delete" (as far as I can see), so it tries to do the import as normal.

lib/puppet/provider/importtemplatexml.rb
asaenz: I can do something cleaner like that much easier once we get the BIOS settings changes in (then I should be able to specifically pass IntegratedRaid = Enabled through asm-deployer).

lib/puppet/provider/importsystemconfiguration/default.rb
gavin-scott: Can raid_action delete be gotten rid of in favor of passing in an empty list of virtual disks to create?

lib/puppet/provider/importtemplatexml.rb
gavin-scott: I'd like to hold off on this until after that then if possible.

asaenz:
@gavin-scott I've made some drastic changes to this PR, so you might want to do another review.

lib/puppet/provider/idrac.rb
```diff
@@ -22,7 +22,7 @@ def exists?
       reset
       exporttemplate
     end
-    synced = !resource[:force_reboot] && config_in_sync?
+    synced = resource[:ensure] != :teardown && !resource[:force_reboot] && config_in_sync?
```
gavin-scott: I thought we were leaning toward just using :absent instead of having the custom :ensure. Do you like :teardown better?

gavin-scott: spoke with @asaenz we agree to stick with teardown

PR #110, opened 2015-02-10
Support for advanced bios settings with clone from reference

asaenz:
NEEDS TO BE MERGED WITH:
https://github.com/dell-asm/asm/pull/215 and https://github.com/dell-asm/asm-deployer/pull/760
as well as corresponding asm-core changes


gavin-scott:
+1

PR #111, opened 2015-02-11
Added default bios_settings so teardown doesn't break.

asaenz:


asaenz:
Should be added to the other bios config PR

PR #113, opened 2015-03-17
Bios settings now removed if they don't exist on target

asaenz:
I also slipped in a few Ruby formatting fixes.


gavin-scott:
@asaenz there were some test failures.

asaenz:
@gavin-scott Tests are passing, and I even added a basic test.

PR #117, opened 2015-03-25
skip background configuration in virtual_disks_ready?

gavin-scott:
This change was made in 8.1 branch but never backported to 8.0


gavin-scott:
This is a backport of #103

PR #118, opened 2015-03-26
ASM-3439 wrong exception rescued in firmware retry

gavin-scott:
Should have been ASM::WsMan::Error not ASM::WsMan::StandardError


gavin-scott:
@asaenz can you review this masterwork?

asaenz:
+1

PR #120, opened 2015-03-27
import config should no longer rely on /var/nfs open permissions

asaenz:
This change is for 8.2, and should probably wait until then unless needed beforehand for whatever reasons.

### DIFF
--- a/lib/puppet/provider/exporttemplatexml.rb
+++ b/lib/puppet/provider/exporttemplatexml.rb
@@ -6,12 +6,12 @@ include REXML
 
 class Puppet::Provider::Exporttemplatexml <  Puppet::Provider
 
-  def initialize (ip,username,password, resource, nfswritepath='/var/nfs', name_postfix='original')
+  def initialize (ip,username,password, resource, name_postfix='original')
     @ip = ip
     @username = username
     @password = password
     @resource = resource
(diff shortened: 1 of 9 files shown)


gavin-scott:
@asaenz tests are failing on this. Holding for 8.2 per recommendation.

gavin-scott:
Please rebase yet again.

gavin-scott:
@asaenz if you have time to rebase this we can get it merged.

asaenz:
@gavin-scott This has been rebased for up to date code.  Sorry I kept putting this off, but whenever we want to get this change in (I assume 8.2, since it's a bit late for 8.1.1)...it should be ready.

gavin-scott:
Hold for 8.2

asaenz:
Accidentally deleted the branch this was on (restored it though now).  Reopening.

asaenz:
Just a note....
We'll need to make a change in asm-core here as well https://github.com/dell-asm/asm-core/blob/master/lib/libcommons/WSManageLib2/src/main/java/com/dell/pg/orion/wsman/compound/ApplyexportXmlConfig.java#L54
The asm-core change doesn't need to be made at the same time as this, but needs to be made before we change /var/nfs permissions.
Mostly leaving the note here for reference.

PR #121, opened 2015-04-02
ASM-3534 fix sporadic idrac firmware reboot job failures

gavin-scott:
The Tempfile being passed to the idrac reboot job was being
garbage collected away in some cases. Change to use the common
ASM::WsMan.reboot method.

### DIFF
--- a/lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
+++ b/lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
@@ -4,6 +4,7 @@ require 'nokogiri'
 require 'erb'
 require 'tempfile'
 require 'asm/util'
+
 require File.join(provider_path, 'idrac')
 
 Puppet::Type.type(:idrac_fw_installfromuri).provide(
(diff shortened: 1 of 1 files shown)


gavin-scott:
Deployed changes on @sushilrai 's appliance, he is running a test.

logankimmel:
@gavin-scott ****Don't Merge this.  Will break all firmware updates

lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
```diff
@@ -113,8 +114,8 @@ def update(firmware_list)
     reboot_id = nil
     if reboot_required
       if @force_restart
-        reboot_config_file_path = create_reboot_config_file
-        reboot_id = create_reboot_job(reboot_config_file_path)
+        require 'asm/wsman'
```
logankimmel: This needs to return the reboot_id for us to check it below or will fail

lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
```diff
@@ -113,8 +114,8 @@ def update(firmware_list)
...
+        ASM::WsMan.reboot(transport, Puppet)
```
logankimmel: reboot_id = ASM::WsMan.reboot(transport, Puppet)

gavin-scott:
Updated with @logankimmel 's fix, but it will require a change to the ASM::WsMan.reboot method to return the job id:

https://github.com/dell-asm/asm-deployer/blob/cb7ce03890c7bde157d2018e2b19fb4067a82e14/lib/asm/wsman.rb#L118

DO NOT MERGE! Requires the above to be changed in asm-deployer

logankimmel:
@gavin-scott Let's talk about this.  I don't think the proposed solution is going to work for several reasons.

gavin-scott:
@logankimmel pointed out to me that the ASM::WsMan.reboot method won't work because it initiates the job. This module needs to bundle it up with the other firmware update jobs before initiating.

PR #122, opened 2015-04-03
ASM-3534 TempFile garbage collect fix

logankimmel:
Return the object instead of the path, this will protect from a
file being garbage collected too soon


gavin-scott:
+1

gavin-scott:
Had success with this on @sushilrai 's appliance.

PR #124, opened 2015-04-09
ASM-3590 Now more correctly applying read only bios settings when given

gavin-scott:


gavin-scott:
8.1.0 backport of https://github.com/dell-asm/dell-idrac/pull/123

PR #125, opened 2015-04-13
Small changes on BIOS settings checks.

asaenz:
1) Now checking target server's BIOS Enumeration if we should apply the bios setting from the reference config
2) Value for 'none' has been changed to n/a from the Java side


asaenz:
Should be merged with: https://github.com/dell-asm/asm-core/pull/1391

PR #126, opened 2015-04-14
Support for none w/ Raid Config option in target boot device

asaenz:
IMPORTANT: Merge with https://github.com/dell-asm/asm-deployer/pull/860 and https://github.com/dell-asm/asm-core/pull/1396

PR #128, opened 2015-04-16
ASM-3603 improve import from reference sever reliability

gavin-scott:
- only apply bios settings if they appear in BIOS enumeration
- support for target boot type of none w/ raid configuration
- remove BiosBootSeq from xml if boot type of none
- add clear lc job queue after idrac reset


gavin-scott:
8.1.0 backport of #125, #126, #127

PR #129, opened 2015-04-16
Clear job queue improvements

asaenz:
Job queue will now be cleared before we do a reboot of the server instead of after.  We will also query the LC to check that the job queue is also empty before allowing it to go on

I also fixed the tests.  yay.

### DIFF
--- a/lib/puppet/provider/importsystemconfiguration/default.rb
+++ b/lib/puppet/provider/importsystemconfiguration/default.rb
@@ -84,9 +84,9 @@ Puppet::Type.type(:importsystemconfiguration).provide(
       reset
       Puppet.info("Waiting for Lifecycle Controller be ready")
       lcstatus
+      clear_job_queue
       reboot
       lcstatus
-      clear_job_queue
     end
     exporttemplate('base')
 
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/importsystemconfiguration/default.rb
```diff
@@ -122,9 +122,25 @@ def clear_job_queue
...
   end
 
-  def sleep_time
-    5
+  def wait_for_jobs_clear
+    Puppet.info("Waiting for job queue to be empty...")
+    endpoint={:host => transport[:host], :user => transport[:user], :password => transport[:password]}
+    schema = "http://schemas.dell.com/wbem/wscim/1/cim-schema/2/DCIM_JobService"
+    10.times do
+      resp = ASM::WsMan.invoke(endpoint, 'enumerate', schema)
```
gavin-scott: you could probably pass in the :selector argument to do the parsing for you, i.e. something like:

    n_jobs = ASM::WsMan.invoke(endpoint, 'enumerate', schema, :selector => '//n1:CurrentNumberOfJobs', :logger => Puppet)

Also might want to pass in logger as in example above.

But code should work as is.

PR #130, opened 2015-04-16
ASM-3603 Clear job queue improvements

gavin-scott:
Job queue will now be cleared before we do a reboot of the server instead of after.  We will also query the LC to check that the job queue is also empty before allowing it to go on


gavin-scott:
8.1.0 backport of https://github.com/dell-asm/dell-idrac/pull/129

PR #134, opened 2015-04-21
ASM-3630 ASM-3634 Import system config read issues

gavin-scott:


gavin-scott:
8.1.0 backport of #133

PR #135, opened 2015-04-22
Now removing HddSeq in all cases

asaenz:


gavin-scott:
@asaenz please include jira in title and include link to PR on jira ticket.

PR #136, opened 2015-04-22
ASM-3657 Now removing HddSeq in all cases

gavin-scott:


gavin-scott:
8.1.0 backport of https://github.com/dell-asm/dell-idrac/pull/135

PR #138, opened 2015-04-22
ASM-3659 Removed extra import execution for reference config imports

gavin-scott:


gavin-scott:
8.1.0 backport of #137

PR #139, opened 2015-04-22
ASM-3660 Boot device of none was not removing invalid partitions

asaenz:


gavin-scott:
This looks like it not maintain the previous behavior where if user does not specify any networks for boot type of none we will pass their NIC config straight through.

PR #140, opened 2015-04-24
ASM-3683 Added wait for running jobs to finish before import/export

asaenz:


lib/puppet/idrac/util.rb
```diff
@@ -36,6 +36,29 @@ def self.virtual_disks_ready?
         end
         true
       end
+
+      def self.wait_for_running_jobs
+        require 'asm/wsman'
+        transport = get_transport
+        Puppet.info("Checking for running jobs")
+        10.times do
+          endpoint={:host=>transport[:host], :user=>transport[:user], :password=>transport[:password]}
```
gavin-scott: transport itself can be passed as the endpoint, no need for extra variable

PR #141, opened 2015-04-27
ASM-3658 BIOS settings from UI now overridden by service necessary settings

asaenz:


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -127,16 +127,24 @@ def get_config_changes
...
+        if @bios_settings[key] == 'n/a'
+          changes['remove']['attributes']['BIOS.Setup.1-1'] ||= []
+          changes['remove']['attributes']['BIOS.Setup.1-1'] << key
+        else
+          changes['partial']['BIOS.Setup.1-1'][key] = @bios_settings[key]
+        end
+      end
+    end
+    unless @resource[:target_boot_device].start_with?('none')
+      changes['partial']['BIOS.Setup.1-1']['ProcVirtualization'] = 'Enabled'
```
gavin-scott: @aimers1975 do we always need ProcVirtualization to be enabled? Or only with hyper-v / esxi?

PR #143, opened 2015-04-28
ASM-3534 TempFile garbage collect fix

gavin-scott:
Return the object instead of the path, this will protect from a
file being garbage collected too soon

change reboot type to 1

Forceful reboot


gavin-scott:
8.1.0 backport of https://github.com/dell-asm/dell-idrac/pull/122

PR #142, opened 2015-04-27
ASM-3661 RAID sections now removed if not hd/none with hd target boot device

asaenz:

PR #144, opened 2015-04-29
ASM-3661 RAID sections now removed if not hd/none with raid target boot

logankimmel:


logankimmel:
810 backport for https://jira.labs.dell.com/browse/ASM-3661
from: #142

PR #145, opened 2015-04-30
ASM-3701 VirtIscsiMacAdrr was sometimes being set as empty string

asaenz:


gavin-scott:
@asaenz what's the jira number on this?

asaenz:
@gavin-scott Added the jira number to the title

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -372,12 +370,21 @@ def munge_iscsi_partitions(nc, changes)
   end
 
   def munge_virt_mac_addr(nc, changes)
+    virtMacAddr = ''
+    virtIscsiMacAddr = ''
```
gavin-scott: Can these be removed? Looks like they are unconditionally set below before they are used.

gavin-scott: Sounds like we need this initialization, but should be moved into the partition loop below.

asaenz:
@gavin-scott Made the change to move the initialization into the loop

PR #147, opened 2015-04-30
Check if VLanMode is valid before trying to set on NICS

gavin-scott:


gavin-scott:
8.1.0 backport of #146

PR #148, opened 2015-04-30
ASM-3682 All exports should now reset idrac if it fails

asaenz:


gavin-scott:
¿Dónde está mi jira?

asaenz:
@gavin-scott It's in the title

lib/puppet/provider/idrac.rb
```diff
@@ -226,23 +220,33 @@ def setup_idrac
...
-          transport[:password],
-          resource,
-          '/var/nfs',
-          postfix
-      )
-      obj.exporttemplatexml
+    begin
+      execute_export_config(postfix)
+    rescue
+      Puppet.debug 'Export template failed.'
```
gavin-scott: Can you add the exception to the debug message?

asaenz:
@gavin-scott Added the exception to the debug message as requested

PR #150, opened 2015-05-03
ASM-3701 VirtIscsiMacAdrr was sometimes being set as empty string

gavin-scott:


gavin-scott:
8.1.0 backport of #145

PR #151, opened 2015-05-03
ASM-3682 All exports should now reset idrac if it fails

gavin-scott:


gavin-scott:
8.1.0 backport of #148

PR #152, opened 2015-05-03
ASM-3660 Boot device=none was importing invalid nic config in some cases

gavin-scott:


gavin-scott:
8.1.0 backport of https://github.com/dell-asm/dell-idrac/pull/139

PR #155, opened 2015-05-08
ASM-3529 Added  new import to fix iscsi/fcoe offload not being set

asaenz:
There were also some changes to improve the code a bit 
1) We now wait for import inside of importtemplatexml.rb rather than having the object pass back the job id and waiting outside.  This is just to "compartmentalize" the import process so the code is kept more "contained"
2) Wrapped the different import calls with the retry code and cleaned that up a bit.
3) Various improvements to the setup process.  Now, we should be checking now if anything needs to change before we actually do the import.

IMPORTANT: Should be merged with https://github.com/dell-asm/asm/pull/234

### DIFF
--- a/lib/puppet/idrac/util.rb
+++ b/lib/puppet/idrac/util.rb
@@ -6,6 +6,7 @@ require 'pty'
 module Puppet
   module Idrac
     module Util
+      class ConfigError < Exception; end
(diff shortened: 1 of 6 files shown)


jkrauskopf:
@sushilrai Please review, thanks.

PR #161, opened 2015-07-06
ASM-4110 Add retry to job que setup for fw_update

logankimmel:


lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
```diff
@@ -276,12 +276,20 @@ def create_reboot_job(reboot_file)
...
+    3.times do |t|
+      resp = run_wsman(wsman_cmd)
+      doc = Nokogiri::XML(resp)
+      if doc.xpath('//n1:MessageID').text == 'SUP025'
+        Puppet.debug("Job Queue created successfully")
+        break
+      else
+        message = doc.xpath('//n1:Message').text
+        if message.include? 'Job cannot be scheduled' && t < 3
+          Puppet.debug("Error scheduling Job Queue.  ..retrying")
```
gavin-scott: We should probably sleep here a bit before retrying.

Also, were you able to get any info from the LC logs as to why the job could not be scheduled? I wonder if it's because the LC is not "ready". If that's the root cause we could wait for that instead of just doing a naive sleep by calling this code:

https://github.com/dell-asm/dell-idrac/blob/b9dd5412b46125ec7c1033ed0c3ad33b129069b2/lib/puppet/provider/idrac.rb#L24

lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
```diff
@@ -276,12 +276,20 @@ def create_reboot_job(reboot_file)
...
-      raise Puppet::Error, "Problem scheduling the job queue.  Message: #{doc.xpath('//n1:Message').text}"
+    3.times do |t|
+      resp = run_wsman(wsman_cmd)
+      doc = Nokogiri::XML(resp)
+      if doc.xpath('//n1:MessageID').text == 'SUP025'
+        Puppet.debug("Job Queue created successfully")
+        break
+      else
+        message = doc.xpath('//n1:Message').text
+        if message.include? 'Job cannot be scheduled' && t < 3
```
gavin-scott: t is always less than 3, that clause could be removed.

This code works fine, but this is a nice pattern for retry that @ripienaar pointed out:

https://github.com/dell-asm/asm-deployer/blob/master/lib/asm/service_deployment.rb#L3816-L3829

Note line 3816 there is actually unnecessary because nil.to_i.succ => 1, but maybe it's better for clarity...

lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
logankimmel: I went ahead and added a sleep here.  Also, we are already checking for LC ready on https://github.com/dell-asm/dell-idrac/blob/master/lib%2Fpuppet%2Fprovider%2Fidrac_fw_installfromuri%2Fwsman.rb#L171

PR #162, opened 2015-07-21
ASM-4331 use local ip for import / export by default

gavin-scott:
If nfsipaddress is not specified in the importsystemconfiguration
resource, use the current IP by default. This allows the same
configuration to function properly on any worker node.


sushilrai:
+1

PR #166, opened 2015-07-29
ASM-3406 remove asm_decrypt password encryption

gavin-scott:
An encrypted device password was passed to the asm-deployer device
management service and written to the device configuration file
url. That password has been deprecated and all puppet-layer code
should use the credential_id argument in preference to that.


gavin-scott:
Note that the importraidconfiguration resource this affects is not actually being used anywhere....

PR #169, opened 2015-08-03
ASM-4213 Import uses graceful shutdown, and only forces if we can't do that

asaenz:


jkrauskopf:
@sushilrai This is ready for review, thanks.

PR #171, opened 2015-08-20
ASM-4431 HyperV iSCSI failure when 57800 is skipped by ASM

sushilrai:
Skipping the import configuration operation on the FQDDs for which MAC address information is not accessible. We are skipping the MAC address information for NIC 3 and 4 for 57800. This change will help in skipping the configuration for these NICS. The attributes set for other NPAR NIC configuration is not applicable to 1GB NICs from 57800 NIC Card

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -646,6 +646,8 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
         partitioned = interface['partitioned']
         interface.partitions.each do |partition|
           fqdd = partition.fqdd
+          # Skipping the FQDDs where MAC address information is not avaialble.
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -646,6 +646,8 @@ def process_nics
         partitioned = interface['partitioned']
         interface.partitions.each do |partition|
           fqdd = partition.fqdd
+          # Skipping the FQDDs where MAC address information is not avaialble.
+          next if partition.mac_address.nil?
```
gavin-scott: I don't think this is right. There is a case where the mac address is expected to be nil -- when we are going from not partitioned to partitioned.

gavin-scott:
Closing, I don't think this is needed.

PR #172, opened 2015-09-03
ASM-4683  Issue with scheduling job queue

logankimmel:
Per email from Tex, the OS_Collector is also a direct update,
we were trying to schedule this and it was failing.  Also, Idrac
and Lifecycle should be run together first before anything else

### DIFF
--- a/lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
+++ b/lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
@@ -15,9 +15,10 @@ Puppet::Type.type(:idrac_fw_installfromuri).provide(
   LC_ID = 28897
   UEFI_DIAGNOSTICS_ID = 25806
   DRIVER_PACK = 18981
+  OS_COLLECTOR = 101734
 
-  # Component ids that do not require a reboot
-  NO_REBOOT_COMPONENT_IDS = [IDRAC_ID, LC_ID, UEFI_DIAGNOSTICS_ID, DRIVER_PACK]
+  # Component ids that do not require a reboot (DIRECT UPDATES)
+  NO_REBOOT_COMPONENT_IDS = [IDRAC_ID, LC_ID, UEFI_DIAGNOSTICS_ID, DRIVER_PACK, OS_COLLECTOR]
 
   # Max time to wait for a job to complete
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
```diff
@@ -34,13 +35,10 @@ def create
     sleep 20
     pre = []
     main = []
-    post = []
     @firmwares.each do |firmware|
       Puppet.debug(firmware)
-      if firmware["component_id"].to_i == LC_ID
+      if firmware["component_id"].to_i == (LC_ID || IDRAC_ID)
```
gavin-scott: This doesn't work to match either LC_ID or IDRAC_ID, just checks whether it's equal to LC_ID. You would need something like `[LC_ID, IDRAC_ID].include?(firmware["component_id"].to_i)`

lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
```diff
@@ -50,14 +48,7 @@ def create
...
     update(main)
-    if post.size > 0
-      Puppet.debug("IDRAC update required, installing last")
-      update(post)
-
-      # idrac restarts after the firmware is installed. Sleep for a minute to
-      # ensure that is underway before we check for LC ready below.
-      sleep(60)
-    end
+    sleep(60)
```
gavin-scott: Do we only need to do this if IDRAC_ID was installed? I think leaving in a comment about the rationale would be a good idea.

lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
logankimmel: How do you figure?
```
irb(main):001:0> a = 1
=> 1
irb(main):002:0> b = 2
=> 2
irb(main):004:0> 1 == (a || b)
=> true
irb(main):005:0> 3 == (a || b)
=> false
```

logankimmel:
@gavin-scott This has been updated.

gavin-scott:
@logankimmel verified that this does indeed fix the issue where a job repeatedly can't be scheduled.

PR #174, opened 2015-09-21
ASM-4683  Issue with scheduling job queue

shaorong-zhu:
Per email from Tex, the OS_Collector is also a direct update,
we were trying to schedule this and it was failing.  Also, Idrac
and Lifecycle should be run together first before anything else

(cherry picked from commit a327c5508437d53d9fad88826318b07c1f1e5c9e)
https://jira.labs.dell.com/browse/ASM-4683


gavin-scott:
+1

PR #175, opened 2015-09-25
ASM-4629 Added retry to wait for pending/committed changes to clear

asaenz:


asaenz:
@gavin-scott I think https://github.com/dell-asm/asm-deployer/pull/1096 will fix the tests.
I took out that gemspec because asm-deployer was no longer being used as a dependency.  But since I had to readd the dependency to this module, I still need the gemspec to be available for pulling in here.

PR #181, opened 2016-01-20
ASM-5595 return value of ASM::WsMan.get_power_state changed

gavin-scott:


gavin-scott:
@asaenz I'm not aware that this has been causing failures, but it had the same problem as https://github.com/dell-asm/asm-deployer/pull/1307 here.

asaenz:
It probably just sent a command to power it on even if was already powered on.  Or maybe the deployment was never in the state where it was already powered on where that would have caused errors.
:+1:

PR #183, opened 2016-02-11
ASM-5753 BIOS VT settings

logankimmel:
Don't hardcode BIOS VT settings to enabled when specifying target
boot device


jkrauskopf:
@gavin-scott or @sushilrai Please review, thanks.

PR #184, opened 2016-02-12
ASM-5561 Server Puppet module changes for Phase 1 vSAN support for RA HCL

sushilrai:
VSAN supporting changes for iDRAC. Update CurrentControlProperty to HBA / RAID based on VSAN property

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -433,7 +433,7 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
     else
       if @boot_device =~ /VSAN/i
         if target_current_xml.to_s.match(/="CurrentControllerMode">RAID/)
-          changes['whole']['RAID.Integrated.1-1'] = { 'CurrentControllerMode' => "HBA"}
+          changes['whole'][raid_configuration.keys.first] = { 'CurrentControllerMode' => "HBA"}
         end
       elsif @boot_device =~ /WITH_RAID|HD/i
         changes['partial'] = {'BIOS.Setup.1-1'=> {'HddSeq' => raid_configuration.keys.first}} if @boot_device =~ /HD/i
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -167,6 +167,13 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
       changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'InternalSdCard' => 'On'})
       changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'HddSeq' => 'Disk.SDInternal.1-1'})
     end
+
+    if @boot_device =~ /VSAN/i
+      changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'InternalSdCard' => 'On'})
+      changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'IntegratedRaid' => 'Enabled'})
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -167,6 +167,13 @@ def default_changes
       changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'InternalSdCard' => 'On'})
       changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'HddSeq' => 'Disk.SDInternal.1-1'})
     end
+
+    if @boot_device =~ /VSAN/i
+      changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'InternalSdCard' => 'On'})
+      changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'IntegratedRaid' => 'Enabled'})
+      changes['partial'].deep_merge!('BIOS.Setup.1-1' => {'HddSeq' => 'Disk.SDInternal.1-1'})
```
gavin-scott: Above seems to assume the server always has integrated raid and a Disk.SDInternal.1-1 I'm guessing the IntegratedRaid setting would just be discarded later if we see that the target server doesn't actually have integrated raid, but the HddSeq setting would probably cause a failure if Disk.SDInternal.1-1 did not exist.

It would seem like we need to be calling controller_views and physical_disk_views to find out which controllers and disks are available.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -423,7 +431,11 @@ def get_raid_config_changes(target_current_xml)
       Puppet.debug("Setting RAID configuration to be cleared as part of teardown.")
       raid_configuration.keys.each{|controller| changes['whole'][controller] = { 'RAIDresetConfig' => "True" } }
     else
-      if @boot_device =~ /WITH_RAID|HD/i
+      if @boot_device =~ /VSAN/i
+        if target_current_xml.to_s.match(/="CurrentControllerMode">RAID/)
+          changes['whole']['RAID.Integrated.1-1'] = { 'CurrentControllerMode' => "HBA"}
```
gavin-scott: Same comment as above re assumption of integrated vs. slot RAID.

lib/puppet/provider/importtemplatexml.rb
sushilrai: We need to take care of this at the server filtering level where, for VSAN configuration  servers to Integrated RAID and SD Card is selected.

gavin-scott: @AlexDron / @aimers1975 can you confirm that server-side filtering is required _integrated raid_ and an SD disk in position `1-1` for the VSAN feature?

gavin-scott: Spoke with @aimers1975 , it's doubtful that we're filtering for this. Generally if you picked e.g. a RAID deployment with a certain number of SD disks we are just validating that the server has a RAID controller (not specifically integrated raid) and that it has the required number of SD disks (not which slots they are in).

AlexDron: We certainly don't do that. I was never told to check it.

gavin-scott: I think I'm confused here, based on @aimers1975 's e-mail Disk.SDInternal.1-1 refers to the SD card. I thought it referred to an SSD. And the IntegratedRaid setting should be disregarded if it doesn't exist on the server, so that's not a problem either.

gavin-scott:
:+1:

lib/puppet/provider/importtemplatexml.rb
gavin-scott: My mistake, please ignore @AlexDron

PR #186, opened 2016-03-07
ASM-953 force all server deployments with SD card(s) to use Mirror redundancy mode

ksw2599:
force all server deployments with SD card(s) to use Mirror redundancy mode


ksw2599:
@gavin-scott can you review/merge please?

gavin-scott:
You were able to test that this works if the server only has 1 SD card I assume?

ksw2599:
yeap. the installation completes, but the server component will show error state. I talked @aimers1975 about this, and she said it's ok. The error comes from idrac itself complaining about the missing flash media. However, the installation and operation seems to work as normal - just without backup write.

PR #188, opened 2016-03-08
ASM-5892 Pending systemconfig failures

logankimmel:
If ImportSystemConfig continues to have pending system configuration,
after a JID_CLEARALL and retry, issue a JID_CLEARALL_FORCE

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -69,10 +69,22 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
         retry
       end
     rescue Puppet::Idrac::PendingChangesError
+      pending_attempts += 1
       if pending_attempts <= 10
-        pending_attempts += 1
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -69,11 +69,24 @@ def execute_import(file_name=@resource['configxmlfilename'])
         retry
       end
     rescue Puppet::Idrac::PendingChangesError
+      pending_attempts += 1
       if pending_attempts <= 10
         pending_attempts += 1
```
gavin-scott: This needs to come out since you moved it to line 72.

gavin-scott:
Looks good to me other than the one comment. @asaenz  can you review as well? The goal here is to run the undocument `JID_CLEARALL_FORCE` command iff there appear to be no jobs running but ImportSystemConfiguration seems to be failing with the pending config error.

asaenz:
:+1:

PR #190, opened 2016-03-18
ASM-6183 Always disable ErrPrompt

asaenz:


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -153,48 +153,51 @@ def get_config_changes
...
+          changes["remove"]["attributes"]["BIOS.Setup.1-1"] << key
         else
-          changes['partial']['BIOS.Setup.1-1'][key] = @bios_settings[key]
+          changes["partial"]["BIOS.Setup.1-1"][key] = @bios_settings[key]
         end
       end
     end
+
+    # Always disable ErrPrompt so we can avoid a hangup due to a user needing to put input on the server manually
+    changes["partial"]["BIOS.Setup.1-1"]["ErrPrompt"] = "Disabled"
```
gavin-scott: Is this the only real change? Everything else is just `'` changed to `"`?

gavin-scott:
I think this is supposed to go into 8.2.0 -- can you reissue this against release/dell-asm-8.2.0?

Other than that :+1:

lib/puppet/provider/importtemplatexml.rb
asaenz: Yeah, this line is the only substantial change.

asaenz:
Oh, whoops. I'll fix it. Though I'll leave this PR since we probably want it for both 8.2 and 8.3, yeah?

asaenz:
8.2 PR: https://github.com/dell-asm/dell-idrac/pull/191

gavin-scott:
The plan is to try to merge from 8.2 branch back into master periodically so that we have to do less manual cherry-picks. So should be able to do just 8.2 branch for 8.2 fixes and just master for 8.3 changes. My understanding is that it won't hurt to go ahead and merge this into master though, so I'll do that.

PR #198, opened 2016-03-29
ASM-5482 Fixing the tests for windows. 

scottcsims:
ASM-5482  Fixing the tests for windows.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,13 +1,24 @@
 source 'https://rubygems.org'
 
+
 gem 'activesupport'
 gem 'nokogiri', '1.5.10'
 gem 'dell-asm-util', :git => 'https://github.com/dell-asm/dell-asm-util.git', :branch => 'master'
 
+# Add gems necessary to run facter on Windows
+platforms :mswin, :mingw do
+  gem 'sys-admin'
+  gem 'win32-process'
+  gem 'win32-dir'
+  gem 'win32-security'
+  gem 'win32-service'
+  gem 'win32-taskscheduler'
+  gem 'windows-pr'
+end
+
 group :development, :test do
-  gem 'asm-deployer', :git => 'git@github.com:dell-asm/asm-deployer.git'
-  gem 'rake'
-  gem 'rspec'
+  gem 'rake',"~> 10.0"
+  gem 'rspec', '~> 2.14'
   gem 'puppetlabs_spec_helper'
   if puppetversion = ENV['PUPPET_GEM_VERSION']
     gem 'puppet', puppetversion
--- a/lib/puppet/idrac/util.rb
+++ b/lib/puppet/idrac/util.rb
@@ -1,7 +1,7 @@
 require 'rexml/document'
 
 include REXML
-require 'pty'
+
 
 module Puppet
   module Idrac
@@ -10,12 +10,13 @@ module Puppet
     class ShutdownError < Exception; end
     class PendingChangesError < Exception; end
     module Util
-      # FIXME: asm/device_management is dependent on asm-deployer, which should be eliminated as a dependency for this module
+
       def self.get_transport
-        require 'asm/device_management'
-        @transport ||= begin
-          ASM::DeviceManagement.parse_device_config(Puppet[:certname])
-        end
+        @transport
+      end
+
+      def self.transport=(transport)
+        @transport=transport
       end
 
       def self.view_disks(type=:virtual)
--- a/spec/unit/puppet/provider/importtemplatexml_spec.rb
+++ b/spec/unit/puppet/provider/importtemplatexml_spec.rb
@@ -22,7 +22,7 @@ describe Puppet::Provider::Importtemplatexml do
   end
 
   before(:each) do
-    @test_config_dir = URI(File.join(Dir.pwd, "spec", "fixtures"))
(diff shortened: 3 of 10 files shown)


Gemfile
```diff
@@ -1,13 +1,35 @@
 source 'https://rubygems.org'
+ruby "1.9.3"
 
 gem 'activesupport'
 gem 'nokogiri', '1.5.10'
 gem 'dell-asm-util', :git => 'https://github.com/dell-asm/dell-asm-util.git', :branch => 'master'
 
+platforms :ruby, :mswin, :mingw do
+  gem 'pg', '0.17.1'
```
gavin-scott: We shouldn't need pg, shouldn't we get that as a transitive dependency of asm-deployer?

Gemfile
```diff
@@ -1,13 +1,35 @@
 source 'https://rubygems.org'
+ruby "1.9.3"
```
gavin-scott: Do we really want to limit this to 1.9.3? What about if folks outside of ASM want to use it with a newer version of ruby?

lib/puppet/idrac/util.rb
```diff
@@ -1,7 +1,7 @@
 require 'rexml/document'
 
 include REXML
-require 'pty'
+#require 'pty'
```
gavin-scott: Let's just remove this rather than leaving it commented out.

spec/unit/puppet/provider/importtemplatexml_spec.rb
```diff
@@ -22,7 +22,7 @@ def build_nic_views(fqdd_to_mac, vendor = nil, product = nil)
   end
 
   before(:each) do
-    @test_config_dir = URI(File.join(Dir.pwd, "spec", "fixtures"))
+    @test_config_dir = URI( File.expand_path('../../../../fixtures', __FILE__))
```
gavin-scott: Moving forward we're trying to use double-quotes everywhere instead of single quotes. See ruby style guide https://github.com/styleguide/ruby We're starting to enforce this in some projects using rubocop. If existing code is using single quotes its fine to leave that but don't change code from using single quotes to double quotes.

spec/unit/puppet/provider/importtemplatexml_spec.rb
```diff
@@ -355,7 +355,7 @@ def build_nic_views(fqdd_to_mac, vendor = nil, product = nil)
       @network_configuration = JSON.parse(File.read(@test_config_dir.path + '/network_configuration.json'))['networkConfiguration']
       #ASM::NetworkConfiguration.any_instance.stub(:new).and_return(ASM::NetworkConfiguration.new(@network_configuration))
       #ASM::NetworkConfiguration.any_instance.stub(:add_nics!).and_return(nil)
-      @test_config_dir = URI(File.join(Dir.pwd, "spec", "fixtures"))
+      @test_config_dir = URI( File.expand_path('../../../../fixtures', __FILE__))
```
gavin-scott: double quotes

Gemfile
```diff
@@ -1,13 +1,35 @@
...
-  gem 'rspec'
+
+  #private github gems need ssh from travis ci and https from the dev environment
+  gem 'asm-deployer', :git => 'git@github.com:dell-asm/asm-deployer.git', :branch => 'master'
+
+  #For Local Development uncomment this line
+  #gem 'asm-deployer', :git => 'https://github.com/dell-asm/asm-deployer.git', :branch => 'master'
+
+  gem 'rake',"~> 10.0"
+  gem 'rspec', '~> 2.14'
```
gavin-scott: I was able to run `bundle exec rake` on windows without locking the rake and rspec versions. Do we really need to lock those down?

gavin-scott:
Please squash the commits into one and have the title as the first line of the commit message e.g. "ASM-5482 Fixing the tests for windows" like you have as the title of the PR. See guidelines here: https://confluence.gtie.dell.com/display/ASM/Github+PR+and+JIRA+Best+Practices

Gemfile
scottcsims: I was unable to bundle install and run the specs with Ruby 2 and Rspec 3 on windows for this project. I had to use Ruby 1.9.3 p448 and Rspec2. I know that Ruby 1.9.3 p448 installed with RVM with Rspec 2 will work on Linux for this project also.

Gemfile
scottcsims: I think that we should use bundle exec rspec as our measurement. Rake is just going to make sure the Rakefile dependencies are wired up. It won't actually make sure that the library code is available and working like the rspec tests will do.  For example it won't catch instances where our remote gem repos are not pulled down in the Gemfile because the Rake file does not depend on all the gems in the bundle.

gavin-scott: I was using `bundle exec rake` to execute rspec. That's what the default task does:

https://github.com/dell-asm/dell-idrac/blob/310974f72da13f82fd7f8837090bd5bc444e9aa0/Rakefile#L34

Gemfile
gavin-scott: Still want it removed. Folks should be able to use this module on ruby 2.x

scottcsims:
I'm including these changes in another PR for ASM-5482 with all of the completed code.

PR #199, opened 2016-03-30
ASM-6335 Power server off after system config applied

gavin-scott:
This prevents the server from booting into who-knows-what is installed
on it after the configuration is complete. Server should be manually
powered on separately when needed.


gavin-scott:
Meant to issue against 8.2 branch.

PR #200, opened 2016-03-30
ASM-6335 Power server off after system config applied

gavin-scott:
This prevents the server from booting into who-knows-what is installed
on it after the configuration is complete. Server should be manually
powered on separately when needed.


gavin-scott:
See ASM-6335 for further rationale as to why we want the server to be off afterward.

gavin-scott:
@asaenz please review.

asaenz:
Discussion with @gavin-scott led to the conclusion of possibly adding a power off to asm-deployer as well to ensure that even if we don't execute ImportSystemConfiguration (rare, but it can happen sometimes), that the server will be powered off when we configure the boot order.

gavin-scott:
Addressed comment with https://github.com/dell-asm/asm-deployer/pull/1514

asaenz:
:+1:

PR #206, opened 2016-04-18
Enabling integrated RAID for ASM-6441 after tearing down 'SD without …

Shreya99:
…RAID' deployment

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -177,6 +177,10 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
     if @boot_device == 'SD'
       changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"IntegratedRaid" => "Disabled"})
     end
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -177,6 +177,10 @@ def default_changes
     if @boot_device == 'SD'
       changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"IntegratedRaid" => "Disabled"})
     end
+    #Enabling raid after 'SD without RAID' deployment
+    if @boot_device == 'SD' && @resource[:ensure] == :teardown
```
gavin-scott: I would remove this block and just change at line 177 to be:

    if @boot_device == 'SD' && @resource[:ensure] != :teardown

gavin-scott:
Test failures are due to recent changes in asm-deployer that are causing rbvmomi to get required. @scottcsims is working on changes to pull the asm-deployer dependency out of this module which should fix it. Will hold this PR until that is in.

PR #207, opened 2016-04-20
Re-enabling integrated-raid after tearing down the 'SD without raid' d…

Shreya99:
…eployment


gavin-scott:
This is a re-issue of #206 

@Shreya99 as we mentioned in stand-up yesterday please do not close PRs and reopen them if you just want to update them. We lose comment history that way. See this wiki for more details: https://confluence.gtie.dell.com/display/ASM/Github+PR+and+JIRA+Best+Practices

gavin-scott:
Tests passing now that #208 has been merged removing the asm-deployer dependency.

PR #208, opened 2016-04-20
ASM-5482 removing the asm-deployer dependency

scottcsims:
Changing where the idrac module gets user,password,ip_address info for the transport method so that we don't depend on asm-deployer.
This needs dell-asm/asm#290

PR #210, opened 2016-04-25

sushilrai:
Added support for boot_device "VSAN_AHCI" where Embedded SATA mode will be set to AHCIMode and SATA Disk 1 will be added as first disk in HDD Sequence. This will enable OS installation of first SATA Disk

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -184,7 +184,13 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
       changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"HddSeq" => "Disk.SDInternal.1-1"})
     end
 
-    if @boot_device =~ /VSAN/i
+    if @boot_device =~ /AHCI_VSAN/i
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"EmbSata" => "AhciMode"})
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"SecurityFreezeLock" => "Disabled"})
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"WriteCache" => "Disabled"})
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"HddSeq" => get_first_sata_disk})
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"InternalSdCard" => "Off"}) if is_sd_card?
+    elsif @boot_device =~ /VSAN/i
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -184,7 +184,12 @@ def default_changes
       changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"HddSeq" => "Disk.SDInternal.1-1"})
     end
 
-    if @boot_device =~ /VSAN/i
+    if @boot_device =~ /AHCI_VSAN/i
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"EmbSata" => "AhciMode"})
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"SecurityFreezeLock" => "Disabled"})
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"WriteCache" => "Disabled"})
+      changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"HddSeq" => "Disk.SATAEmbedded.A-1"})
```
gavin-scott: Is it always Disk.SATAEmbedded.A-1? @AlexDron does your filtering for AHCI_VSAN guarantee that the server has this disk?

AlexDron: Yes, AHCI_VSAN is a target boot type for this kind of filtering

gavin-scott: I guess my question was -- when you filter for AHCI_VSAN are we guaranteed that the server has a Disk.SATAEmbedded.A-1? I think the answer is no based on when I reviewed the filtering PR. If we're not guaranteed of that then we need to either add that as part of filtering or add code here that enumerates the disks and gets the first one present.

sushilrai:
@gavin-scott  I am working on the logic for getting the SATA disk enumeration. Have sent query to the LC team on this.

The tests are failing due to rbvmomi package require. I dont see any reference of rbvmomi in idrac module. Even my changes are very limited to the boot order and havent added changes which add the dependency for rbvmomi.
Can you please take a look at it.

gavin-scott:
@sushilrai if you rebase the tests should be fixed. The rbvmomi issue was fixed by #208 which removed the need for the idrac module to use asm-deployer code. (that's what was pulling in rbvmomi)

sushilrai:
@gavin-scott  this PR is merged to master branch but sill the changes are missing the 8.2.0 branch. Are we planning to port it back to 8.2.0. Need this PR to complete VSAN changes for Urgent Fury work

gavin-scott:
Tests are failing due to asm-deployer dependency that we can't get rid of for 8.2.0. I'll go ahead and merge.

PR #211, opened 2016-04-27
Revert "ASM-6005 Extend idrac waits"

gavin-scott:
Reverts dell-asm/dell-idrac#209


gavin-scott:
This was breaking deployments as the `wait_for_lc_ready` method is in a different class.

PR #215, opened 2016-05-04

sushilrai:
To ensure we are picking the first disk in the enumeration for OS installation


sushilrai:
@gavin-scott please review the change, The tests are failing due to asm-deployer dependency.

aimers1975:
@sushilrai @gavin-scott is this change needed and what are we doing with it?  We need all VSAN related work in before Sunday...

PR #216, opened 2016-05-11
ASM-6656 Use Wsman library for firmware update

logankimmel:
Use WsMan library for running firmware update commands in order
to properly handle escaping of special characters and errors


logankimmel:
Requires: https://github.com/dell-asm/dell-asm-util/pull/46

logankimmel:
Looks like tests are failing due to another dependency

PR #217, opened 2016-05-17
Replaced ',' with '&' in dell-idrac module

Shreya99:


gavin-scott:
Closing in favor of #228

PR #218, opened 2016-05-17

sushilrai:
Skip HDD Sequence from input XML in case requested value is already set. This is added to take care of Config BIOS Job remaining in the running state for ever in case existing and current value is same.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -268,6 +268,22 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
         @changes['partial']['BIOS.Setup.1-1'].delete('BiosBootSeq')
       end
     end
+
+    existing_hdd_seq = find_hdd_seq(xml_base)
+    hdd_seq_change = @changes['partial']['BIOS.Setup.1-1']['HddSeq']
+    Puppet.debug("Existing HDD SEQ: #{existing_hdd_seq}")
+    Puppet.debug("HDD Seq Change: #{hdd_seq_change}")
+    if existing_boot_seq && boot_seq_change
+      seq_diff = hdd_seq_change.delete(' ').split(',').zip(existing_hdd_seq.delete(' ').split(',')).select{|new_val, exist_val| new_val != exist_val}
+      #If tearing down, the HDD will already be removed from the boot sequence
+      if seq_diff.size ==0 || @resource[:ensure] == :teardown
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -369,6 +385,25 @@ def find_bios_boot_seq(xml_base)
     nil
   end
 
+  #Helper function which just searches through the xml comments for HddSeq value, since it will be commented out
+  def find_hdd_seq(xml_base)
```
gavin-scott: Instead of copy-pasting `find_bios_boot_seq` can you rename it something like `find_bios_attribute(xml_base, attr_name)` and pass in the attribute you are looking for? Either BiosBootSeq or HddSeq.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -268,6 +268,22 @@ def munge_config_xml
...
+    hdd_seq_change = @changes['partial']['BIOS.Setup.1-1']['HddSeq']
+    Puppet.debug("Existing HDD SEQ: #{existing_hdd_seq}")
+    Puppet.debug("HDD Seq Change: #{hdd_seq_change}")
+    if existing_hdd_seq && hdd_seq_change
+      seq_diff = hdd_seq_change.delete(' ').split(',').zip(existing_hdd_seq.delete(' ').split(',')).select{|new_val, exist_val| new_val != exist_val}
+      #If tearing down, the HDD will already be removed from the boot sequence
+      if seq_diff.size ==0 || @resource[:ensure] == :teardown
+        @changes['partial']['BIOS.Setup.1-1'].delete('HddSeq')
+      end
+      if existing_hdd_seq.split(",").first.strip == hdd_seq_change
```
gavin-scott: What does this block do? Seems like this case should have been caught by `seq_diff.size == 0` above. The code this was copied from doesn't do this:

https://github.com/dell-asm/dell-idrac/blob/b869a24708d811ac67d48c6bdab2f2b23b6e8663/lib/puppet/provider/importtemplatexml.rb#L270-L276

aimers1975:
@gavin-scott is this a change that needs to go in with the rest of the VSAN work?  Looks like maybe it is not good to merge?  Is Sushil going to have to look at this tonight?

asaenz:
👍 Though it would be lovely if we could squash these commits into a single useful one.

gavin-scott:
Tests are failing due to asm-deployer dependency pulling in rbvmomi. Fixed in 8.3

PR #221, opened 2016-05-19
ASM-6719 Import configuration operation failing for VSAN with SD card in case HBA mode is not already set

sushilrai:
Import configuration operation failing for VSAN with SD card in case HBA mode is not already set. This loop never got tested as we were using the servers which were atleast once configured to be used as VSAN node.
This bug got introduced while addressing the code comments. After addressing the code comments this part was not tested.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -460,8 +460,7 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
             raid_configuration = {}
             disk_types.keys.each do |disk|
               controller = disk.split(':').last
-              raid_configuration[controller] ||= []
-              raid_configuration[controller] << disk
+              raid_configuration[controller][:hotspares] << disk
             end
             Puppet.debug("Inside VSAN RAID Configuration: #{raid_configuration}")
           elsif !(unprocessed['virtualDisks'].empty? && unprocessed['externalVirtualDisks'].empty?)
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -464,7 +464,7 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
               raid_configuration[controller] << disk
             end
             Puppet.debug("Inside VSAN RAID Configuration: #{raid_configuration}")
-          elsif unprocessed['virtualDisks'].empty? && unprocessed['externalVirtualDisks'].empty?
+          elsif !(unprocessed['virtualDisks'].empty? && unprocessed['externalVirtualDisks'].empty?)
             (unprocessed['virtualDisks'] + unprocessed['externalVirtualDisks']).each do |config|
               type = disk_types[config['physicalDisks'].first]
               #Just check first disk in the list to get what type of virtual disk it is
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -456,7 +456,15 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
             disk_types[fqdd] = type
           end
           raid_configuration = Hash.new { |h, k| h[k] = {:virtual_disks => [], :hotspares => []} }
-          unless unprocessed['virtualDisks'].empty? && unprocessed['externalVirtualDisks'].empty?
+          if unprocessed.nil? && @boot_device.match(/VSAN/i)
+            raid_configuration = {}
+            disk_types.keys.each do |disk|
+              controller = disk.split(':').last
+              raid_configuration[controller] ||= []
+              raid_configuration[controller] << disk
+            end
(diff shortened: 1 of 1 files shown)


gavin-scott:
@asaenz can you review?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -456,7 +456,15 @@ def raid_configuration
             disk_types[fqdd] = type
           end
           raid_configuration = Hash.new { |h, k| h[k] = {:virtual_disks => [], :hotspares => []} }
-          unless unprocessed['virtualDisks'].empty? && unprocessed['externalVirtualDisks'].empty?
+          if unprocessed.nil? && @boot_device.match(/VSAN/i)
```
gavin-scott: Can you paste in a snippet of the ImportSystemConfiguration block that this generates? I think that will make it clearer to me what this does.

aimers1975:
@sushilrai can you please address comments...

lib/puppet/provider/importtemplatexml.rb
sushilrai: In case of 'SD Card with RAID' input parameters do not have raid_configuration in the input resource as a result "unprocessed" has is nil.

For VSAN we need to set CurrentControllerMode to HBA for RAID Controller

`<Component FQDD="RAID.Integrated.1-1">
    <Attribute Name="CurrentControllerMode">HBA</Attribute>
  </Component>
`

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -456,7 +456,15 @@ def raid_configuration
...
+            raid_configuration = {}
+            disk_types.keys.each do |disk|
+              controller = disk.split(':').last
+              raid_configuration[controller] ||= []
+              raid_configuration[controller] << disk
```
asaenz: I don't completely like this just because it makes our raid_configuration object not be consistent. In this case, raid_configuration will be something like {"RAID.Integrated.1-1" => "Disk-Fqdd.1-1"}, whereas usually it'll look something like {"RAID.foo.1-1" => {:virtual_disks => {:disks => ["Disk-Fqdd.1-1"], :level => ...}, :hotspares => {...}}.

I understand this is a quick fix and probably okay overall, but I dislike how the data structure isn't consistent between these cases. At the very least, there should be more comments so this behavior is documented.

Just for the sake of consistency, it would make more sense and be a little clearer if you just did something like raid_configuration[controller] = {} or even {:virtual_disks =. {}, :hotspares => {}} to keep the object consistent. I think this would help to avoid any potential issues of accidentally hitting another piece of code that processes those.

gavin-scott: Yeah, I wondered why the disks were being added, it seems like they are not being used. Can you update per @asaenz 's suggestion?

sushilrai:
@ripienaar please review and merge the changes

ripienaar:
tests failing @sushilrai

sushilrai:
@ripienaar three tests are failing due to dependency of rbvmomi in 8.2.1 branch. multiple PRs are merged with these tests failing.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -456,7 +456,13 @@ def raid_configuration
             disk_types[fqdd] = type
           end
           raid_configuration = Hash.new { |h, k| h[k] = {:virtual_disks => [], :hotspares => []} }
-          unless unprocessed['virtualDisks'].empty? && unprocessed['externalVirtualDisks'].empty?
+          if unprocessed.nil? && @boot_device.match(/VSAN/i)
+            disk_types.keys.each do |disk|
+              controller = disk.split(':').last
+              raid_configuration[controller][:hotspares] << disk
```
gavin-scott: I would have preferred we didn't add the disks at all since we're not actually going to add these disks as hot spares to a RAID volume. The whole concept of using the `raid_configuration` method just to get controller information seems very questionable. There's a DCIM_ControllerView we could use instead to get that. Unfortunately dell-asm-util had a typo in the URI, I just submitted a fix for that: https://github.com/dell-asm/dell-asm-util/pull/54 With that you get:

    [root@dellasm ~]# wsman_shell.rb --server 172.17.12.60
    [1] pry(#<ASM::WsMan>)> controller_views.map {|e| e[:fqdd] }
     => ["RAID.Integrated.1-1", "AHCI.Embedded.1-1", "AHCI.Embedded.2-1"]

gavin-scott:
Re failing tests -- if I'd known we'd have so many more 8.2 changes I probably would have recommended we just add rbvmomi to the Gemfile to fix them even though this module doesn't really use that library. Do you want to add that in on the new PR #223 ?

ripienaar:
@gavin-scott would be nice, but probably not worth it unless we're addressing your comments above about disks?

gavin-scott:
It looks like there's at least one more fix that has to go in for 8.2.1 as @sushilrai added #223 , so could add it to that.

PR #223, opened 2016-05-20
ASM-6720 Config XML import operation fails to set HBA mode when Virtual Disks are present

sushilrai:
Included "RAIDresetconfig" attribute with the RAID controller while moving the Controller from RAID to HBA mode. This will ensure all the Virtual Disks are removed before the HBA mode is configured.
In case there is any Virtual Disk on the RAID controller, then HBA mode cannot be set.

Also, then Controller is already in HBA mode then RAIDresetconfig is not allowed by iDRAC that why we are limiting the RAIDresetconfig attribute only in case we are moving controller from RAID to HBA mode.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -497,6 +497,8 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
       if @boot_device =~ /VSAN/i
         if target_current_xml.to_s.match(/="CurrentControllerMode">RAID/)
           changes['whole'][raid_configuration.keys.first] = { 'CurrentControllerMode' => "HBA"}
+          # Need to remove Virtual Disks when we move Controller from RAID to HBA mode.
+          raid_configuration.keys.each{|controller| changes['whole'][controller] = { 'RAIDresetConfig' => "True" } }
         end
       elsif @boot_device =~ /WITH_RAID|HD/i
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -497,6 +497,8 @@ def get_raid_config_changes(target_current_xml)
       if @boot_device =~ /VSAN/i
         if target_current_xml.to_s.match(/="CurrentControllerMode">RAID/)
           changes['whole'][raid_configuration.keys.first] = { 'CurrentControllerMode' => "HBA"}
+          # Need to remove Virtual Disks when we move Controller from RAID to HBA mode.
+          raid_configuration.keys.each{|controller| changes['whole'][controller] = { 'RAIDresetConfig' => "True" } }
```
gavin-scott: I'm confused. I thought the whole problem the PR was addressing was that `raid_configuration` was failing because `unprocessed` was nil here:

https://github.com/dell-asm/dell-idrac/blob/f983180f585d449ee4a12f99fc0c11ff14b8f583/lib/puppet/provider/importtemplatexml.rb#L465

gavin-scott:
Oh, I see this is maybe a fix for the previous PR that got merged #221 ?

sushilrai:
Yes @gavin-scott  this is another bug which I found yesterday and we wanted to include this in the build as part of patch if everything works fine with this change.

gavin-scott:
I'll merge this and will file a jira to clean up the `raid_configuration` stuff in 8.3

PR #227, opened 2016-06-01
ASM-6781 Fix transport in idrac_fw_installfromuri

gavin-scott:
8.2 changes that were made to use the ASM::WsMan library in
idrac_fw_installfromuri to fix password-handling problems weren't
compatible with 8.3 changes to use a transport resource.

### DIFF
--- a/lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
+++ b/lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
@@ -27,6 +27,7 @@ Puppet::Type.type(:idrac_fw_installfromuri).provide(
 
 
   def exists?
+    transport # Force initialization, Puppet::Idrac::Util depends on it
     @force_restart = resource[:force_restart]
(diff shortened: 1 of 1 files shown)


gavin-scott:
@logankimmel can you review?

logankimmel:
I don't see where we are ever closing the transport/ wsman_client.  Is this an issue?

gavin-scott:
It's a bit funky but the `transport` returned from the parent provider is really just an encapsulation of the endpoint data so it doesn't need to be closed:

https://github.com/dell-asm/dell-idrac/blob/fdaca7f89353c57502dc0f3243da421aebea5ea2/lib/puppet/provider/idrac.rb#L151


https://github.com/dell-asm/dell-idrac/blob/fdaca7f89353c57502dc0f3243da421aebea5ea2/lib/puppet_x/puppetlabs/transport/idrac.rb#L13-L19

logankimmel:
That's right, sorry.  +1

PR #228, opened 2016-06-05
ASM-6558 Interoperability with wsmancli 2.6

gavin-scott:
Update ws-man URLs to use "&" as query parameter separators; wsmancli
2.6 requires this. These URLs with work earlier wsmancli versions such
as 2. 3 as well.


gavin-scott:
@asaenz please review.

asaenz:
👍

PR #229, opened 2016-06-15
ASM-6942 VSAN iDRAC configuration fails when RAID controller mode is not in HBA Mode

sushilrai:
When RAID Controller is not in HBA mode, we have to set it to HBA and RAIDReset Attribute needs to be passed to ensure all the Virtual Disk and Hotspare configuration is removed.
Earlier code was trying to set the RAIDReset for all the RAID controllers which is not supported by iDRAC also initialization of the variable was not correct.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -496,9 +496,8 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
     else
       if @boot_device =~ /VSAN/i
         if target_current_xml.to_s.match(/="CurrentControllerMode">RAID/)
-          changes['whole'][raid_configuration.keys.first] = { 'CurrentControllerMode' => "HBA"}
-          # Need to remove Virtual Disks when we move Controller from RAID to HBA mode.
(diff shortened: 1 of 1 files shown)


sushilrai:
@gavin-scott please review and merge the changes.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -496,9 +496,8 @@ def get_raid_config_changes(target_current_xml)
     else
       if @boot_device =~ /VSAN/i
         if target_current_xml.to_s.match(/="CurrentControllerMode">RAID/)
-          changes['whole'][raid_configuration.keys.first] = { 'CurrentControllerMode' => "HBA"}
-          # Need to remove Virtual Disks when we move Controller from RAID to HBA mode.
-          raid_configuration.keys.each{|controller| changes['whole'][controller].merge!({ 'RAIDresetConfig' => "True" }) }
+          changes['whole'][raid_configuration.keys.first] = { 'CurrentControllerMode' => "HBA",
+                                                              'RAIDresetConfig' => "True" }
```
gavin-scott: I'll go ahead and merge this since it's

PR #231, opened 2016-06-17
ASM-6520 iDRAC discovery scripts

kaushalgala:
The files here are copies of the idrac discovery scripts in asm module.

Once this change is merged, we will remove the idrac discovery scripts from asm


gavin-scott:
I don't want to move this here because it introduces a dependency on the private asm-deployer code. @AlexDron -- if we need to move this here we will need to change the AsmManager/ServerDiscoveryRequest and JRAF/DiscoveryRequest APIs to accept a plaintext username and password instead of a credential id. That way this script can just use the username and password and not have to do the ASM credential decryption.

kaushalgala:
Discard this PR. Don't merge

PR #233, opened 2016-06-29
Removed RAID.Embedded.1.1 component from input XML for VSAN deployment on back-plane disk

sushilrai:
- Current code was looking for the first RAID component and in certain cases RAID.Embedded component was getting picked instead of RAID.Integrated

- Removed RAID.Embedded component from the input XML in case installation disk in AHCI Mode


asaenz:
👍

PR #234, opened 2016-07-06
ASM-7035 VSAN deployment fails with VSAN partition on back side disk

sushilrai:
Run import configuration twice to take care of removing backplane raid configuration. With 13G servers there is no way to initialize the disk in AHCI and ATA mode. For now we are deleting the VD when server is in Raidmode and then update the server from Raid mode to AHCI mode

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -19,11 +19,45 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
     @templates_dir = File.join(Puppet::Module.find('idrac').path, 'templates')
     @exported_postfix = exported_postfix
     @boot_device = resource[:target_boot_device]
+    @import_operation_count = 1
+    @ahci_mode = false
   end
 
   def importtemplatexml
-    munge_config_xml unless @xml_processed
-    execute_import
+    if @boot_device == "AHCI_VSAN"
+      2.times do
+        Puppet.debug("Import Count: #{@import_operation_count}")
+        Puppet.debug("Boot Device: #{@boot_device}")
+        munge_config_xml
+        execute_import
+
+        break if @ahci_mode
+
+        @xml_base = nil
+        @xml_processed = false
+        @changes = nil
+        @import_operation_count += 1
+        export_config
+      end
+    else
+      munge_config_xml unless @xml_processed
+      execute_import
+    end
+  end
+
+  def export_config(name_postfix='original')
+    export_file_name = File.basename(@resource[:configxmlfilename], ".xml")+ "_#{name_postfix}.xml"
+    Puppet::Idrac::Util.wait_or_clear_running_jobs
+    require 'asm/util'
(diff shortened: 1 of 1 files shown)


aimers1975:
@sushilrai is it expected that the tests are failing.. @asaenz are you able to review despite the failing tests.  Note this is an urgent 8.2.1 fix.

sushilrai:
@ripienaar not sure which gem version has changed. Tests are failing due the version conflict which require ruby 2.0

ripienaar:
@sushilrai looking, but yeah its not obvious whats going on...

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -311,8 +390,8 @@ def munge_config_xml
   def remove_invalid_settings(xml_to_edit)
     # HddSeq seems to cause a lot of issues by letting it stay.
     # We only allow it to stay if we're specifically trying to set it for booting from hard drive
-    hdd_seq = find_attribute_value(xml_to_edit, 'BIOS.Setup.1-1', 'HddSeq', false)
-    hdd_seq.remove unless hdd_seq.nil? || @changes['partial']['BIOS.Setup.1-1']['HddSeq']
+    #hdd_seq = find_attribute_value(xml_to_edit, 'BIOS.Setup.1-1', 'HddSeq', false)
```
asaenz: I can't 100% remember the issues that these lines solved, but why is this being commented out?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -19,11 +19,45 @@ def initialize (ip,username,password,resource, exported_postfix='base')
...
+        @import_operation_count += 1
+        export_config
+      end
+    else
+      munge_config_xml unless @xml_processed
+      execute_import
+    end
+  end
+
+  def export_config(name_postfix='original')
```
asaenz: I don't like how much of this is copy/pasted from exporttemplatexml.rb. Couldn't you just use the exporttemplatexml object similar to here?  https://github.com/dell-asm/dell-idrac/blob/master/lib/puppet/provider/idrac.rb#L165

lib/puppet/provider/importtemplatexml.rb
sushilrai: "hdd_seq" is a string variable and string.remove is not a valid operation.

asaenz: Oh. Wow, that's been in for quite a while. I wonder how that got past me...weird. Alright then.

ripienaar:
@sushilrai i fixed the gem install thign

lib/puppet/provider/importtemplatexml.rb
gavin-scott: The issue they were trying to solve is the one where if you don't actually change the HddSeq or BiosBootSeq but leave it set to an existing value the idrac import config job hangs forever.

gavin-scott:
@sushilrai do you think this PR is still needed? @aimers1975 reports it did not seem to solve the original problem.

gavin-scott:
Closing, reopen if still needed.

PR #235, opened 2016-07-12
ASM-7008 Added support for putting disks into nonraid mode

asaenz:
Importsystemconfiguration wsman action does not support changing raid
disk modes, so this functionality was added in a separate puppet type
using direct wsman commands.

This new type could potentially be used for any other raid configuration
workflows that importsystemconfiguration does not support, or that we
want to do before importsystemconfiguration happens.


asaenz:
MUST BE MERGED WITH https://github.com/dell-asm/asm/pull/314

asaenz:
DO NOT MERGE YET, still testing

PR #239, opened 2016-07-27
ASM-7185 Win2008R2 Bare-Metal Install Failing with Error "O/S install  timed out"

sushilrai:
Disk initialization XML output has changed. As a result we were not waiting for the disk initialization to complete that was leading to intermittent Windows OS installation failure. This has been observed with Windows 2008R2 very consistently and occasionally with Windows 2012 / HyperV deployments.

### DIFF
--- a/lib/puppet/idrac/util.rb
+++ b/lib/puppet/idrac/util.rb
@@ -47,9 +47,9 @@ module Puppet
 
       def self.virtual_disks_ready?
         response = view_disks
-        response.xpath('//DCIM_VirtualDiskView').each do |disk|
+        response.xpath('//DCIM_VIRTUALDiskView').each do |disk|
           current_op = disk.at_xpath('//OperationName').content
-          unless current_op =~ /None|Background/
+          if current_op =~ /Background/
             fqdd = disk.at_xpath('FQDD').content
             percent = disk.at_xpath('OperationPercentComplete').content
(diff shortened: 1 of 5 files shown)


ripienaar:
@sushilrai tests failing, do the same JSON thing here.  IN THIS PR, dont make a new one. then this will jsut go green

lib/puppet/idrac/util.rb
```diff
@@ -49,7 +49,7 @@ def self.virtual_disks_ready?
         response = view_disks
         response.xpath('//DCIM_VirtualDiskView').each do |disk|
           current_op = disk.at_xpath('//OperationName').content
-          unless current_op =~ /None|Background/
+          if current_op =~ /Background/
```
ripienaar: we dont need to worry about backward compat?

sushilrai: It seems code was always skipping the check. We have to look for pattern "Background" for OperationName. @gavin-scott can comment more on this.

ripienaar: OK - I don't know this stuff, so will leave for @gavin-scott to merge

gavin-scott: I thought the Background initialization can last many many hours and we were purposefully allowing disks that are in Background. Why are you changing this?

gavin-scott:
Alternate fix already merged: https://github.com/dell-asm/razor-server/pull/108 Closing...

PR #240, opened 2016-07-29
ASM-7383 Teardown of non-RAID disks

logankimmel:
If non-RAID disks exist, on teardown we need to convert
these to a RAID virtual disk so we can re-use the server


logankimmel:
@gavin-scott  This is good to go. Tested it on my appliance.  There's no teardown tests in this module currently so in the interest of time I submitted it without. Can add some after the fact but I'm confident in the logic here

logankimmel:
Tests failing due to bundler issue. Not quite sure how to fix this

PR #241, opened 2016-08-01
ASM-7423 Deployment teardown throws exception in logs

kaushalgala:
The deployment teardown did not have RAID config specified in deployment.json.
As a result, asm::idrac had set raid_configuration to {}, which later on
results in nil class exception in the puppet module.

The fix is to have safety nil checks before invoking empty? method


gavin-scott:
I think the RAID configuration is empty when the user has specified the static VSAN configuration. My only concern here is that we may need to be doing some other teardown operations for that configuration.

@sushilrai do you know if we are doing any RAID teardown for VSAN currently? Seems like we need to be.

kaushalgala:
@gavin-scott Based on sushil's note, there is no RAID teardown applicable. So it might make sense to have the safety checks in this PR.

PR #243, opened 2016-08-11
[DNM] ASM-7468 Move Idrac reset to the utility module

logankimmel:


ripienaar:
just making a note here: needs https://github.com/dell-asm/dell-asm-util/pull/67

lib/puppet/idrac/util.rb
```diff
@@ -151,25 +151,7 @@ def self.wait_for_jobs_clear
...
-            elsif stream == :stderr
-              raise Puppet::Error, 'Error resetting Idrac'
-            end
-          end
-        end
-        wait_for_idrac(360)
+        ASM::Racadm.new(get_transport, :logger => Puppet).reset_idrac
       end
 
       def self.wait_for_idrac (timeout = 180, state = 0)
```
gavin-scott: Do you want to delete this method out (and make sure any callers are using ASM::Util instead)?

logankimmel:
@gavin-scott updated this with the idrac wait.  I think this is good instead of calling the util directly all the time so we can stub out the method easily

gavin-scott:
My original comment was whether we wanted to remove the `wait_for_idrac` method entirely from this module and replace it's usage with the `ASM::Util` one.

None of these changes are really needed for 8.3 it's just code clean-up; I'm starting to think we should hold this for the next release cycle.

gavin-scott:
Holding until next release cycle.

logankimmel:
@gavin-scott  I'm assuming we can close this?

gavin-scott:
Yep, you're right, closing.

PR #244, opened 2016-08-15
ASM-7582 Do not create VD without any disks

logankimmel:
Conditional was incorrect causing us to create the VD with an
empty set of disks


logankimmel:
Here is how the raid_configuration looks with nonraid disks:
```
{"RAID.Integrated.1-1"=>
  {:virtual_disks=>[{:disks=>["Disk.Bay.1:Enclosure.Internal.0-1:RAID.Integrated.1-1"], :level=>"raid0", :type=>:hdd}],
   :hotspares=>[],
   :nonraid=>["Disk.Bay.0:Enclosure.Internal.0-1:RAID.Integrated.1-1"]}}
```

PR #245, opened 2016-08-15
ASM-7561: removed boot Hddseq if it is already set in idrac

arunrordell:
Previously Idrac job runs forever if the same boot sequesnce is set again. In this PR it removes the Hddseq if it is already set.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -335,8 +335,10 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
   def remove_invalid_settings(xml_to_edit)
     # HddSeq seems to cause a lot of issues by letting it stay.
     # We only allow it to stay if we're specifically trying to set it for booting from hard drive
-    hdd_seq = find_attribute_value(xml_to_edit, 'BIOS.Setup.1-1', 'HddSeq', false)
-    hdd_seq.remove unless hdd_seq.nil? || @changes['partial']['BIOS.Setup.1-1']['HddSeq']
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -335,8 +335,8 @@ def munge_config_xml
   def remove_invalid_settings(xml_to_edit)
     # HddSeq seems to cause a lot of issues by letting it stay.
     # We only allow it to stay if we're specifically trying to set it for booting from hard drive
-    hdd_seq = find_attribute_value(xml_to_edit, 'BIOS.Setup.1-1', 'HddSeq', false)
-    hdd_seq.remove unless hdd_seq.nil? || @changes['partial']['BIOS.Setup.1-1']['HddSeq']
+    hdd_seq = find_attribute_value(xml_to_edit, 'BIOS.Setup.1-1', 'HddSeq', true)
+    @changes['partial']['BIOS.Setup.1-1'].delete('HddSeq') if @changes['partial']['BIOS.Setup.1-1']['HddSeq'].eql?(hdd_seq)
```
gavin-scott: Rather than just checking `if @changes['partial']['BIOS.Setup.1-1']['HddSeq'].eql?(hdd_seq)` I think we need to check whether the _first_ entry in the HddSeq is already set to what we want it to be. See e.g. this code:

https://github.com/arunrordell/dell-idrac/blob/bf48c62737b255c4f9c1ba12fbaedd1655d2bb08/lib/puppet/provider/importtemplatexml.rb#L279

I.e. the case we need to cover is where we want to set HddSeq = "RAID.Integrated.1-1", but HddSeq is currently "RAID.Integrated.1-1, Disk.SDInternal.1-1".  In this case we do not want to set HddSeq because RAID is already first.

PR #250, opened 2016-09-28
ASM-7503 Fix teardown NilClass exception when there are no RAID disks

kaushalgala:
Problem:
In the raid_configuration, if there are no RAID config to process (like
SATADOM) the unprocessed hash may be empty, resulting into any looked up
value in this hash as nil e.g. unprocessed['virtualDisks']. Accessing
methods from looked up value results into nil class exception.

Solution:
The check for determining "no RAID config needed" needs to include empty
hash check along with nil value check.


kaushalgala:
@gavin-scott @ripienaar @sushilrai  can you review and merge ?

PR #251, opened 2016-09-30
ASM-7744 Support for RAID on S130

logankimmel:
Allow deployments to S130 embedded RAID card when that card is selected.
If the Embedded controller is currently in a mode other than "RaidMode"
we will have to configure that first before we can do any other configurations.
This will result in 2 passes of the import system config

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -24,7 +24,7 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
   def importtemplatexml
     unless embsata_in_sync?
       Puppet.debug("Embedded Sata settings out of sync. Need to configure first.")
-      munge_config_xml(true)
+      munge_config_xml(:embsata_out_of_sync? => true)
       execute_import
     end
     munge_config_xml unless @xml_processed
@@ -285,14 +285,21 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
     xml_doc.xpath('/SystemConfiguration').first
   end
 
-  def munge_config_xml(emb_mode_change=false)
+  # Munge config xml
+  #
+  # This builds out the appropriate xml file for the requested configuration changes
+  #
+  # @param [Hash] opts
+  # @option opts [Boolean] :embsata_out_of_sync?  True will cause raid configuration to be skipped
+  # @return [Nokogiri::XML::Document] xml_doc
+  def munge_config_xml(opts={})
     get_config_changes
     xml_base.xpath("//Component[contains(@FQDD, 'NIC.') or contains(@FQDD, 'FC.')]").remove unless @changes['whole'].find_all{|k,v| k =~ /^(NIC|FC)\./}.empty?
     xml_base['ServiceTag'] = @resource[:servicetag]
 
     handle_missing_devices(xml_base, @changes)
     @nonraid_to_raid = false
-    @changes.deep_merge!(get_raid_config_changes(xml_base)) unless emb_mode_change
+    @changes.deep_merge!(get_raid_config_changes(xml_base)) unless opts[:embsata_out_of_sync?]
 
     %w(BiosBootSeq HddSeq).each do |attr|
       existing_attr_val = find_bios_attribute(xml_base, attr)
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -22,6 +22,11 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
   end
 
   def importtemplatexml
+    unless embsata_in_sync?
+      Puppet.debug("Embedded Sata settings out of sync. Need to configure first.")
+      munge_config_xml(true)
+      execute_import
+    end
     munge_config_xml unless @xml_processed
     execute_import
   end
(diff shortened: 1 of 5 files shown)


spec/unit/puppet/provider/importtemplatexml_spec.rb
```diff
@@ -402,6 +402,35 @@ def build_nic_views(fqdd_to_mac, vendor = nil, product = nil)
...
+      @fixture.instance_variable_set(:@resource, resource)
+      expect(@fixture.embsata_in_sync?).to eq(true)
+    end
+
+    it "should return false when embSata Raid required and not current" do
+      @fixture.stub(:is_embedded_raid?).and_return(true)
+      expect(@fixture.embsata_in_sync?).to eq(false)
+    end
+  end
    context "when getting SATA disk for boot" do
```
kaushalgala: let's add a space here

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -898,5 +906,27 @@ def find_attribute_value(xml, component, attribute, search_comments=false)
     end
     attr_node.nil? ? nil : attr_node.content
   end
+
+  # Returns true if this is a deployment to an embedded raid (S130) controller
+  #
+  # @return Boolean
+  def is_embedded_raid?
+    virtual_disks = @resource[:raid_configuration].fetch("virtualDisks", [])
```
kaushalgala: not sure, but should we check if @resource[:raid)configuration] is set ? want to make sure we dont run into nil class exception :)

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -277,14 +285,14 @@ def get_xml(postfix=@exported_postfix)
     xml_doc.xpath('/SystemConfiguration').first
   end
 
-  def munge_config_xml
+  def munge_config_xml(emb_mode_change=false)
```
kaushalgala: I guess passing boolean values makes it bit difficult to guess what it means on the caller side
i.e. what does true mean in munge_config_xml(true)

One way to solve this readability issue is by passing a hash of options to the method.
i.e. 
1. the definition would look like : def munge_config_xml(opts)
2. caller will invoke munge_config_xml({ :emb_mode_change? => true })
3. implementation looks like:  @changes.deep_merge!(get_raid_config_changes(xml_base)) unless opts[:emb_mode_change?]

This is one suggestion I've seen folks using before, you may come up with your own solution to tackle the readability suggestion.

logankimmel: That was my first instinct but didn't want to add a whole options hash just for one variable but I think it's a good idea. I'll make the change

lib/puppet/provider/importtemplatexml.rb
logankimmel: How would a nil class happen with the `#fetch` method?, it raid_configuration is nil, it will return empty list

logankimmel:
@kaushalgala This has been updated. github is making your comment on the options param look like that line hasn't changed but it has

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -277,14 +285,14 @@ def get_xml(postfix=@exported_postfix)
     xml_doc.xpath('/SystemConfiguration').first
   end
 
-  def munge_config_xml
+  def munge_config_xml(opts={})
```
ripienaar: now it's harder to figure out what can go into the opts hash so we need yard

kaushalgala: yeah, let's add a yard doc

lib/puppet/provider/importtemplatexml.rb
kaushalgala: ok, makes sense - then ignore this comment

logankimmel:
@ripienaar @kaushalgala  yard doc added

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -277,14 +285,21 @@ def get_xml(postfix=@exported_postfix)
     xml_doc.xpath('/SystemConfiguration').first
   end
 
-  def munge_config_xml
+  # Munge config xml
+  #
+  # This builds out the approriate xml file for the requested configuration changes
```
kaushalgala: change spelling to "appropriate" :)

kaushalgala:
@logankimmel looks good to me, how about squashing into one commit ?

lib/puppet/provider/importtemplatexml.rb
logankimmel: Fixed the typo

logankimmel:
@kaushalgala Can I squash with a merge between my commits ?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -277,14 +285,21 @@ def get_xml(postfix=@exported_postfix)
     xml_doc.xpath('/SystemConfiguration').first
   end
 
-  def munge_config_xml
+  # Munge config xml
+  #
+  # This builds out the appropriate xml file for the requested configuration changes
+  #
+  # @param [Hash] opts
+  # @option opts [Boolean] :embsata_out_of_sync  True will cause raid configuration to be skipped
```
ripienaar: it's wrong here - missing `?`, think its better this way personally

logankimmel: fixed

kaushalgala:
@logankimmel i guess I can squash with merge on github.

Waiting on @ripienaar for approval

PR #253, opened 2016-10-07
ASM-7905 Puppet provider for server onboarding

kaushalgala:
Implementation for new type and provider for configuring credentials
and network settings for a given server using racadm commands. We keep
same style of implementation as we did for blades in dell-cmc module,
except for the fact the racadm calls here go directly to the server.
We also clean up and yard doc some execution methods in racadm.rb

Note: Maybe we should later refactor the code to move racadm command
to a common project where eventually all racadm calls for servers go to
single place. Right now some code is unnecessarily duplicated between
racadm.rb in dell-cmc and dell-idrac modules.

### DIFF
--- a/lib/puppet/provider/idrac_racadm.rb
+++ b/lib/puppet/provider/idrac_racadm.rb
@@ -1,15 +1,21 @@
+require "pathname"
 begin
-  require 'puppet_x/puppetlabs/transport'
-  require 'puppet_x/puppetlabs/transport/racadm'
-rescue LoadError
-  require 'pathname' # WORK_AROUND #14073 and #7788
-  mod = Puppet::Module.find('transport', Puppet[:environment].to_s)
-  require File.join mod.path, 'lib/puppet_x/puppetlabs/transport'
-  require File.join mod.path, 'lib/puppet_x/puppetlabs/transport/racadm'
+  require "puppet_x/puppetlabs/transport"
+rescue LoadError # WORK_AROUND #14073 and #7788
+  mod = Puppet::Module.find("transport", Puppet[:environment].to_s)
+  require File.join mod.path, "lib/puppet_x/puppetlabs/transport" if mod
+end
+begin
+  require "puppet_x/puppetlabs/transport/racadm"
+rescue LoadError # WORK_AROUND #14073 and #7788
+  mod = Puppet::Module.find("transport", Puppet[:environment].to_s)
+  begin
+    require File.join mod.path, "lib/puppet_x/puppetlabs/transport/racadm" if mod
+  rescue LoadError  # This would happen in CI scenario, ignore it since we anyways will mock racadm out
+  end
 end
 
 class Puppet::Provider::IdracRacadm <  Puppet::Provider
-
   def local_permissions
     {"Administrator" => "0x1ff", "Operator" => "0x1f3", "ReadOnly" => "0x1", "None" => "0x0"}
   end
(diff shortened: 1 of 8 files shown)


kaushalgala:
Still working on unit tests, as seem to be having difficulties with dependency module on transport for my tests.  In meanwhile, please review the code and provide feedbacks

lib/puppet/provider/idrac_racadm.rb
```diff
@@ -27,40 +27,68 @@ def client
...
+  # @param subcommand [String] racadm subcommand to invoke
+  # @param params [Array<String>] list of params to be appended to racadm set <fqdd>.<group>.<object>.<index>
+  # @param process_opts [Hash] extra options for processing results
+  # @option process_opts [Boolean] :raise_on_err flag to indicate whether to raise exception if command errs
+  # @return [Object] output for the command result. Can be either a string, or list of strings
+  # @raise [StandardError] when racadm command fails and :raise_on_err flag is set
+  def racadm_cmd(subcommand, params=[], process_opts={})
+    cmd = "racadm %s" % subcommand
+    unless params.empty?
+      param_string = params.is_a?(Array) ? params.join(" ") : params.to_s
```
gavin-scott: `Array(params)` will do the "if array return the array otherwise make an array" check for you. So this block could be e.g.:

    cmd << " %s" % Array(params).join(" ") unless params.empty?

lib/puppet/provider/idrac_racadm.rb
```diff
@@ -27,40 +27,68 @@ def client
...
+      cmd << " %s" % param_string
     end
     output = client.exec!(cmd)
-    Puppet.info("racadm #{subcommand} result: #{output}")
-    Puppet.err("Could not send command racadm #{subcommand}") if output.include?('ERROR:')
+    Puppet.info("racadm %s result: %s" % [subcommand, output])
+    if output.include?("ERROR:")
+      msg = "Error in racadm command %s" % subcommand
+      Puppet.err msg
```
gavin-scott: Please include the parentheses around the arguments: `Puppet.err(msg)`

lib/puppet/provider/idrac_racadm.rb
```diff
@@ -27,40 +27,68 @@ def client
...
+      raise msg if process_opts[:raise_on_err]
+    else
+      Puppet.info "Successfully executed %s" % subcommand
```
gavin-scott: Parentheses

lib/puppet/provider/server_onboard/default.rb
```diff
@@ -0,0 +1,143 @@
...
+  # @param max_wait [Integer] maximum seconds to wait for the IP to connect
+  # @param sleep_interval [Integer] seconds to sleep before re-checking in next interval
+  # @return [void]
+  # @raise [StandardError] when times out waiting for static IP address
+  def wait_for_ip(ip, username, password, max_wait=300, sleep_interval=20)
+    result = Hashie::Mash.new
+    rounds = (max_wait / sleep_interval)
+    connected = false
+
+    cmd = ["sudo", "wsman", "identify", "-h", ip, "-P", "443",
```
gavin-scott: No need to use `sudo` here.

lib/puppet/provider/server_onboard/default.rb
```diff
@@ -0,0 +1,143 @@
...
+           "-u", username, "-p", password, "-c", "dummy.cert", "-y", "basic", "-V", "-v"]
```
gavin-scott: It would probably be a good idea to include `--transport-timeout` here as well. It might be worth adding this "identify" functionality into dell-asm-util's ASM::WsMan class and then we won't have to duplicate the arg setup here.

lib/puppet/type/server_onboard.rb
```diff
@@ -0,0 +1,29 @@
+Puppet::Type.newtype(:server_onboard) do
+  desc "Puppet type to setup networking and new creds on a server through racadm"
+
+  newparam(:name, :namevar => true) do
+    desc "certname of the server"
+  end
+
+  newproperty(:credential) do
```
gavin-scott: It would be better to use `username` and `password` instead of `credential` since the credential is an internal ASM concept.

lib/puppet/type/server_onboard.rb
```diff
@@ -0,0 +1,29 @@
...
+    desc "credential ID to lookup the credentials used to set up admin user"
+  end
+
+  # Network Configuration parameter
+  #
+  # specification for network configuration like static IP, DNS settings
+  #
+  # @param networks [Hash] Options for specifying the network config
```
gavin-scott: An example of this data would be good. I guess that could be an `@example` tag here. Not sure how the YARD stuff will interact with puppet types.

lib/puppet/type/server_onboard.rb
kaushalgala: @gavin-scott to keep it consistent with other places ( e.g. dell-cmc idrac onboard, etc. ) I used credential.  We can discuss on Monday.

lib/puppet/provider/server_onboard/default.rb
kaushalgala: yeah, will move wait_for_ip to dell-asm-util

lib/puppet/type/server_onboard.rb
kaushalgala: ok will do

kaushalgala:
@gavin-scott  addressed all your comments, and also have uploaded my unit tests on the PR.

Like I mentioned in my email, the tests work with rspec, but having trouble with  rake execution. Will need your help on Monday

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: ruby
-script: "bundle exec rake lint && bundle exec rake"
+script: "bundle exec rake lint && bundle exec rake spec"
```
gavin-scott: Should we just change the default task to "spec" instead of "rspec"? e.g. https://github.com/dell-asm/dell-idrac/blob/82a3f2226f9847dece254444bbe6322cbbf90aca/Rakefile#L34 and https://github.com/dell-asm/dell-idrac/blob/82a3f2226f9847dece254444bbe6322cbbf90aca/Rakefile#L40 Otherwise the tests will fail for a developer who just runs "bundle exec rake" correct?

kaushalgala: let me try

lib/puppet/type/server_onboard.rb
```diff
@@ -0,0 +1,62 @@
...
+  #
+  # @option credential [String] "username" username for connecting to the server
+  # @option credential [String] "password" password for connecting to the server
+  #
+  # @example puppet manifest to specify credential
+  #  server_onboard { "rackserver-abcd123" :
+  #    credential = {
+  #      "username" => "foo",
+  #      "password" => "bar"
+  #    }
```
gavin-scott: Nice, looks good.

PR #254, opened 2016-10-13
ASM-8077 Fixed S130 RAID deployments

logankimmel:
Several changes to the xml munging when doing deployments to S130.

* We will not just run the same xml config file up to three times
* Remove the RaidinitConfig setting for S130
* When doing s130 from AHCI,ATA do a full find and replace on the
  xml file to anticipate what the new controller schemas are going
  to be after the embedded sata mode is changed
* Set SecurityFreeze and WriteCache to Disabled
* Do not nest the physical disks within an enclosure

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -22,11 +22,6 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
   end
 
   def importtemplatexml
-    unless embsata_in_sync?
-      Puppet.debug("Embedded Sata settings out of sync. Need to configure first.")
-      munge_config_xml(:embsata_out_of_sync? => true)
-      execute_import
-    end
     munge_config_xml unless @xml_processed
     execute_import
   end
@@ -192,6 +187,8 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
       changes["remove"]["attributes"]["BIOS.Setup.1-1"].push("InternalSdCardRedundancy", "InternalSdCardPrimaryCard")
       if is_embedded_raid?
         changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"EmbSata" => "RaidMode"})
+        changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"SecurityFreezeLock" => "Disabled"})
+        changes["partial"].deep_merge!("BIOS.Setup.1-1" => {"WriteCache" => "Disabled"})
       end
     end
     #Boot Device could be SD_WITHOUT_RAID or SD_WITH_RAID.  Raid Settings are handled above for WITH_RAID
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -289,17 +286,15 @@ def get_xml(postfix=@exported_postfix)
   #
   # This builds out the appropriate xml file for the requested configuration changes
   #
-  # @param [Hash] opts
-  # @option opts [Boolean] :embsata_out_of_sync?  True will cause raid configuration to be skipped
   # @return [Nokogiri::XML::Document] xml_doc
-  def munge_config_xml(opts={})
+  def munge_config_xml()
```
gavin-scott: You can remove the `()` since you removed the args.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -356,7 +351,17 @@ def munge_config_xml(opts={})
...
     # Disable SD card and RAID controller for boot from SAN
-    File.open(@config_xml_path, 'w+') { |file| file.write(xml_base.to_xml(:indent => 2)) }
+
+    File.open(@config_xml_path, 'w+') do |file|
+      if embsata_in_sync?
+        file.write(xml_base.to_xml(:indent => 2))
+      else
+        # If Embedded Sata mode is out of sync we need to change the FQDD's to what they will be
+        # after the EmbSat mode is changed to RAIDmode
+        file.write(xml_base.to_xml(:indent => 2).gsub("AHCI.Embedded", "RAID.Embedded").gsub("ATA.Embedded","RAID.Embedded"))
```
gavin-scott: This sees a little extreme. I would have thought we could just apply this logic when we're processing the controllers and disks.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -597,18 +602,20 @@ def get_raid_config_changes(target_current_xml)
...
+                "Name"              => "ASM VD#{index}",
+                "Size"              => "0",
+                "StripeSize"        => "128",
+                "SpanDepth"         => span_depth,
+                "SpanLength"        => span_length,
+                "RAIDTypes"         => disk_config[:level].sub("raid", "RAID "),
+                "IncludedPhysicalDiskID"=> disk_config[:disks]
+              }
+
+              raid_settings["RAIDinitOperation"] = "Fast" unless is_embedded_raid?
```
gavin-scott: A comment above this line that the S130 doesn't support RAIDinitOperation but that it defaults to "Fast" would be nice.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -929,7 +941,7 @@ def is_embedded_raid?
 
   # Check for Embedded Sata in sync
   #
-  # If the current embedded sata mode is not what is required
+  # If the current embedded sata mode is not what is required +
```
gavin-scott: Stray `+` I guess.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -630,12 +637,17 @@ def set_disk_changes!(disks, type, controller_changes)
...
       disk_attributes = {}
       _bay, enclosure_fqdd = disk_fqdd.split(':', 2)
-      controller_changes[enclosure_fqdd] ||= {}
 
       disk_attributes["RAIDPDState"] = type == :nonraid ? "Non-RAID" : "Ready"
       disk_attributes["RAIDHotSpareStatus"] = "Global" if type == :hotspare
 
-      controller_changes[enclosure_fqdd].merge!({disk_fqdd => disk_attributes})
+      if is_embedded_raid?
+        # Embedded s130 do not have an enclosure
```
gavin-scott: Is this the reason the disks were nested wrong, because the FQDD does not contain an enclosure? What do the FQDDs look like for these disks?

Rather than having to build an `if/else` here could you just do `enclosure_fqdd ||= disk_fqdd` or `enclosure_fqdd = disk_fqdd if is_embedded_raid?`?

logankimmel: This is what caused the nesting problems.  Your solution will not work.  `enclosure_fqdd == disk_fqdd` when using embedded controller

logankimmel:
@gavin-scott  Added tests

lib/puppet/provider/importtemplatexml.rb
gavin-scott: Spoke with @logankimmel and although this is a bit heavy-handed it seems like the safest thing to do for now.

gavin-scott:
Thanks @logankimmel !

PR #255, opened 2016-10-24
ASM-7916 Support FC430 and FC630 VSAN ready nodes on local disk

sushilrai:
Added specific check for All-Flash FC630 servers where need to perform
the OS installation for ESXi on Virtual disk managed for H330 PERC
Mini controller. We need to create a Virtual disk with RAID1
configuration on the disks managed by H330 and use the same for the OS
installation.

Currently there is no specific information is passed from UI which
indicates that we have to use the controller disk for the OS
installation. Thats why some model specific changes has been added


lib/puppet/idrac/util.rb
```diff
@@ -59,6 +59,15 @@ def self.virtual_disks_ready?
         true
       end
 
+      def self.disk_controller
+        output = ASM::WsMan.invoke(get_transport, 'enumerate', "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/DCIM/DCIM_ControllerView")
+        #Response will return multiple base nodes, which will error out with Nokogiri, so we wrap response output
+        xml = Nokogiri::XML("<Envelopes>#{output}</Envelopes>")
+        #Name spaces makes data hard to search through later
+        xml.remove_namespaces!
+        return xml
```
kaushalgala: you can omit return and simply have last line as "xml", as per ruby best practices

kaushalgala: can you also add unit test for this method ?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -948,5 +967,67 @@ def embsata_in_sync?
     return true unless is_embedded_raid?
     find_bios_attribute(original_xml, "EmbSata") == "RaidMode"
   end
+
+  def fc630_with_vsan_on_hdd?
```
kaushalgala: let's not hard code  the method name to the model name, and think of a generic name. What if in future more models require same behavior ?  How about applicable_sata_raid_mode? or something similar that determines characteristics we sought for FC 630 like servers ?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -948,5 +967,67 @@ def embsata_in_sync?
...
+    @resource[:model] == "fc630" && @resource[:target_boot_device] == "AHCI_VSAN"
```
kaushalgala: if possible, we should see if we can avoid hard coding model numbers and look for specific bios attributes. @gavin-scott any idea what would be such check ?

If we end up with still model names, then I would have a list saying APPLICABLE_SATA_RAID_SERVER_MODELS = ["fc630"] and have a include? check here.  That way, in future if we desire same behavior for other models, we just add those model numbers to this list.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -948,5 +967,67 @@ def embsata_in_sync?
...
+      fqdd = x.xpath('FQDD').text
+      Puppet.debug("FQDD: #{fqdd}")
+      disks << fqdd if fqdd =~ /\S+#{device_fqdd}/i
+    end
+    raise("Expect 2 disks, got %d" % [disks.size]) if disks.size != 2
+    disks
+  end
+
+  #TODO: Find disks under H330 controller
+  def fc630_raid_configuration
```
kaushalgala: Just like my previous comment, we should think of a generic name rather than specific model.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -948,5 +967,67 @@ def embsata_in_sync?
...
+      end
+    end
+    return device_id
+  end
+
+  def fc630_controllers
+    ["PERC H330 Mini"]
+  end
+
+  def fc630_disks
```
kaushalgala: Just like my previous comment, we should think of a generic name rather than specific model.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -948,5 +967,67 @@ def embsata_in_sync?
...
+    disk_controllers.xpath('//Envelope/Body/PullResponse/Items/DCIM_ControllerView').each do |x|
+      product_name = x.xpath('ProductName').text
+      if controller_type.include?(product_name)
+        device_id = x.xpath('FQDD').text
+      end
+    end
+    return device_id
+  end
+
+  def fc630_controllers
```
kaushalgala: Just like my previous comment, we should think of a generic name rather than specific model.

lib/puppet/provider/importtemplatexml.rb
gavin-scott: The problem right now is that we have separate code for VSAN and @sushilrai is being given a lot of weird seemingly random requirements for VSAN like "only for FC630". I'd like us to move to having asm-core just populate the correct RAID configuration in the template and pass that through so that we had one path in this module for setting RAID. As opposed to currently where we have specific code for specific model numbers.

But I think for now this is what @sushilrai was asked to deliver.

lib/puppet/idrac/util.rb
gavin-scott: @sushilrai you could just use `ASM::WsMan.controller_views`: https://github.com/dell-asm/dell-asm-util/blob/af992b80efb7288ee25da7152eeb3d903fe97c88/lib/asm/wsman.rb#L111

```
[root@dellasm ~]# wsman_shell.rb --server 172.17.3.210
[1] pry(#<ASM::WsMan>)> controller_views
=> [{:bus=>"1",
  :cache_size_in_mb=>"0",
  :cachecade_capability=>"0",
  :controller_firmware_version=>"25.4.0.0017",
  :device=>"0",
  :device_card_data_bus_width=>"8x or x8",
  :device_card_manufacturer=>"DELL",
  :device_card_slot_length=>"3",
  :device_card_slot_type=>"PCI Express Gen3 x16",
  :device_description=>"RAID Controller in Slot 1",
  :driver_version=>"6.910.18.00",
  :encryption_capability=>"0",
  :encryption_mode=>"0",
  :fqdd=>"RAID.Slot.1-1",
  :function=>"0",
  :instance_id=>"RAID.Slot.1-1",
  :key_id=>nil,
  :last_system_inventory_time=>"20161022090559.000000+000",
  :last_update_time=>"20161022090559.000000+000",
  :max_available_pci_link_speed=>"Generation 3",
  :max_possible_pci_link_speed=>"Generation 3",
  :pci_device_id=>"5F",
  :pci_slot=>"1",
  :pci_sub_device_id=>"1F44",
  :pci_sub_vendor_id=>"1028",
  :pci_vendor_id=>"1000",
  :patrol_read_state=>"1",
  :primary_status=>"1",
  :product_name=>"PERC H330 Adapter",
  :realtime_capability=>"1",
  :rollup_status=>"1",
  :sasaddress=>"544A842041FF6300",
  :security_status=>"0",
  :sliced_vdcapability=>"1",
  :support_controller_boot_mode=>"1",
  :support_enhanced_auto_foreign_import=>"1",
  :support_raid10_uneven_spans=>"1",
  :t10_picapability=>"1"}]
```

lib/puppet/idrac/util.rb
```diff
@@ -59,6 +59,17 @@ def self.virtual_disks_ready?
         true
       end
 
+      def self.disk_controller
+        # output = ASM::WsMan.invoke(get_transport, 'enumerate', "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/DCIM/DCIM_ControllerView")
+        # #Response will return multiple base nodes, which will error out with Nokogiri, so we wrap response output
+        # xml = Nokogiri::XML("<Envelopes>#{output}</Envelopes>")
+        # #Name spaces makes data hard to search through later
+        # xml.remove_namespaces!
+        # return xml
```
gavin-scott: Can you remove the commented out code?

spec/unit/puppet/provider/importtemplatexml_spec.rb
```diff
@@ -613,4 +613,51 @@ def build_nic_views(fqdd_to_mac, vendor = nil, product = nil)
...
+      end
+
+      it "should raise error in no disks are found" do
+        supported_controllers = ["PERC H430 Mini", "PERC H730 Mini"]
+        @fixture.stub(:fc630_controllers).and_return(supported_controllers)
+        message = "Failed to find disks added to controller '%s'" % [supported_controllers]
+        expect { @fixture.fc630_disks }.to raise_error(message)
+      end
+    end
+
```
gavin-scott: delete space

spec/unit/puppet/provider/importtemplatexml_spec.rb
```diff
@@ -613,4 +613,51 @@ def build_nic_views(fqdd_to_mac, vendor = nil, product = nil)
...
+      @test_config_dir = URI( File.expand_path("../../../../fixtures", __FILE__))
+      @controller_disk_xml = eval File.read(@test_config_dir.path + '/controller_view.xml')
+      @physical_disk_view = File.read(@test_config_dir.path + '/physical_disk_view.xml')
+      Puppet::Module.stub(:find).with("idrac").and_return(@test_config_dir)
+      Puppet::Idrac::Util.stub(:disk_controller).and_return(@controller_disk_xml)
+      ASM::WsMan.stub(:invoke).and_return(@physical_disk_view)
+    end
+
+    context "when PERC controller is H330" do
+
```
gavin-scott: delete space

lib/puppet/idrac/util.rb
```diff
@@ -59,6 +59,17 @@ def self.virtual_disks_ready?
...
+        require 'asm/wsman'
```
kaushalgala: minor comment - use double quotes for any new or modified code

lib/puppet/provider/importtemplatexml.rb
kaushalgala: Yeah, we should try to move any custom business logic like this from this module, to asm-deployer under appropriate type / provider schematics, and then let that thing orchestrate desired configuration from dell-idrac module.  We should try to make this dell-idrac module without asm business logic. Anyways it does not seem we can do that now.

@gavin-scott should we file a different future refactor Jira to move such custom business logic out of dell-idrac ?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -948,5 +967,67 @@ def embsata_in_sync?
...
+    ["PERC H330 Mini", "PERC H730 Mini"]
+  end
+
+  def fc630_disks
+    disks = []
+    device_fqdd = controller_disk_fqdd(fc630_controllers)
+    raise("Failed to find disks added to controller '%s'" % [fc630_controllers]) unless device_fqdd
+    physical_disks.xpath('//Envelope/Body/PullResponse/Items/DCIM_PhysicalDiskView').each do |x|
```
kaushalgala: single to double quotes

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -948,5 +967,67 @@ def embsata_in_sync?
...
+      fqdd = x.xpath('FQDD').text
```
kaushalgala: single to double quotes

kaushalgala:
@sushilrai seems tests are failing on:

/home/travis/build/dell-asm/dell-idrac/lib/puppet_x/puppetlabs/transport/racadm.rb:2:in `<top (required)>': uninitialized constant PuppetX (NameError)
	from /home/travis/build/dell-asm/dell-idrac/lib/puppet/provider/idrac_racadm.rb:9:in `require'

sushilrai:
@gavin-scott @kaushalgala Travis test failure is totally unrelated. Can you please take a look into this.

kaushalgala:
@sushilrai seems recent changes in either puppet-transport or dell-asm-util may have caused problems. Looking into it.

aimers1975:
@sushilrai Seems like there is more to do on this PR for you?

gavin-scott:
I rebased and squashed for you @sushilrai  -- tests are re-running now.

kaushalgala:
thanks @gavin-scott

PR #257, opened 2016-10-25
ASM-8143 Embedded RAID Support 2

logankimmel:
There is currently a bug when doing S130 RAID 1 where we cannot
clear the config and create the virtual disk on the same run.
Because of this, we will first run the importtemplate with an
`embedded_change` flag which will skip any raid creation and
only change the EmbMode and tear down raid

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -10,6 +10,8 @@ include REXML
 
 class Puppet::Provider::Importtemplatexml <  Puppet::Provider
 
+  attr_accessor :embedded_sata_change
+
   def initialize (ip,username,password,resource, exported_postfix='base')
     @ip = ip
     @username = username
@@ -294,7 +296,13 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
 
     handle_missing_devices(xml_base, @changes)
     @nonraid_to_raid = false
-    @changes.deep_merge!(get_raid_config_changes(xml_base))
+
+    if embedded_sata_change
+      Puppet.debug("Embedded Mode Change detected running with RAID teardown only")
+      @changes.deep_merge!(get_raid_config_changes(xml_base, raid_reset=true))
+    else
+      @changes.deep_merge!(get_raid_config_changes(xml_base))
+    end
 
     %w(BiosBootSeq HddSeq).each do |attr|
       existing_attr_val = find_bios_attribute(xml_base, attr)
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -294,7 +296,13 @@ def munge_config_xml
 
     handle_missing_devices(xml_base, @changes)
     @nonraid_to_raid = false
-    @changes.deep_merge!(get_raid_config_changes(xml_base))
+
+    if embedded_change
+      Puppet.debug("Embedded Mode Change detected running with RAID teardown only")
+      @changes.deep_merge!(get_raid_config_changes(xml_base, teardown_only=true))
```
gavin-scott: I'm a little confused by the name, maybe `raid_reset=true`

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -562,9 +570,9 @@ def raid_configuration
         end
   end
 
-  def get_raid_config_changes(target_current_xml)
+  def get_raid_config_changes(target_current_xml, raid_reset=false)
     changes = {'partial'=>{}, 'whole'=>{}, 'remove'=> {'attributes'=>{}, 'components'=>{}}}
-    if @resource[:ensure] == :teardown && !@resource[:raid_configuration].nil? && !@nonraid_to_raid
+    if (@resource[:ensure] == :teardown && !@resource[:raid_configuration].nil? && !@nonraid_to_raid) || raid_reset
```
kaushalgala: even with raid_reset, do we need to check !@resource[:raid_configuration].nil? && !@nonraid_to_raid ?

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -562,9 +570,9 @@ def raid_configuration
...
       Puppet.debug("Setting RAID configuration to be cleared as part of teardown.")
```
kaushalgala: This debug log will need to change to reflect teardown or raid_reset flag

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -10,6 +10,8 @@
 
 class Puppet::Provider::Importtemplatexml <  Puppet::Provider
 
+  attr_accessor :embedded_change
```
kaushalgala: probably embedded_sata_change would be more clear, what do you say ?

logankimmel: yeah that's fine. I'll update it

lib/puppet/provider/importtemplatexml.rb
logankimmel: fine

logankimmel:
@kaushalgala updated

kaushalgala:
Thanks @logankimmel  for addressing comments and updating the tests.

@gavin-scott can you take final look and merge ?

PR #258, opened 2016-10-26
ASM-7589 Win2012R2 Deployment Failed with Error "Unable to create new partition or locate existing partition"

sushilrai:
Due to presence of non-raid disk on the server, Virtual disk created for OS installation is not available at "0" disk id. To take care of this we are disabling non-raid disks from the server so that they are not available to the server.
This will also take care of the case where server has non-raid disks created from some prior deployments it is not requested in the existing deployment.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -24,10 +24,6 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
     if fc630_with_vsan_on_hdd?
       @resource[:raid_configuration] = fc630_raid_configuration
     end
-    if !non_raid_disks.empty? && non_raid_not_requested? && !@boot_device.match(/vsan/i)
-      @nonraid_to_raid = true
-      @resource[:raid_configuration]["virtualDisks"] << non_raid_disks
-    end
   end
 
   def importtemplatexml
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -334,7 +338,7 @@ def munge_config_xml
     # If we are tearing down and there are nonraid volumes, we need to make them raid volumes to
     # be able to boot from this controller again
     nonraid_disks = raid_configuration.select{|_,v| !v[:nonraid].empty?}
-    if @resource[:ensure] == :teardown && !nonraid_disks.empty?
+    if (@resource[:ensure] == :teardown && !nonraid_disks.empty?)
```
gavin-scott: Generally for ruby style conformance don't put the parentheses on `if` statements unless you need grouping.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -1036,5 +1049,46 @@ def fc630_raid_configuration
     }
   end
 
+  def non_raid_not_requested?
+    vds = @resource["raid_configuration"]["virtualDisks"]
```
gavin-scott: This could be a little nicer with something like:

```
vds = @resource["raid_configuration"]["virtualDisks"]
return false unless vds
!!vds.find { |vd| vd["raidLevel"] == "nonraid" }
```

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -1036,5 +1049,46 @@ def fc630_raid_configuration
...
+    else
+      return true
+    end
+  end
+
+  def non_raid_disks
+    return @non_raid_info if @non_raid_info
+    disks = []
+    controllers = []
+    physical_disks.xpath("//Envelope/Body/PullResponse/Items/DCIM_PhysicalDiskView").each do |x|
```
gavin-scott: It would be nice to change `physical_disks` to use `ASM::WsMan#physical_disks` so that you don't have to do parsing here. Maybe a bigger change than you want to make now since `self.physical_disks` is used in several other places in this file...

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -1036,5 +1049,46 @@ def fc630_raid_configuration
...
+        "numberofdisks" => "1",
+        "disktype" => "any",
+      },
+      "mediaType" => "ANY"
+    }
+  end
+
 end
 
+
```
gavin-scott: Remove trailing blank lines.

gavin-scott:
Tests are failing.

PR #259, opened 2016-10-28
ASM-8106 SD Deployment with SD Disabled fails

logankimmel:
Changed the logic when gathering the current BiosBootSeq and HddSeq
to actually call out to the Lifecycle controller.
The reason for this being that we were previously comparing what
we wanted to set these attributes to against the original exported config xml
from before we ran any importconfig jobs.  This can sometimes result
in differences between the current config and the exported config
on subsequent import jobs.

If we thought the HddSeq or the BiosBootSeq were out of synce but
actually were not, we would mistakenly try to set this values
and create errors in the lifecycle controller

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -318,9 +318,9 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
     end
 
     %w(BiosBootSeq HddSeq).each do |attr|
-      existing_attr_val = find_bios_attribute(xml_base, attr)
+      existing_attr_val = find_current_boot_attribute(attr.downcase.to_sym)
       requested_val = @changes['partial']['BIOS.Setup.1-1'][attr]
-      message = "Attribute: %s, Existing value: %s, Requested value: %s" % [attr, existing_attr_val, requested_val] 
+      message = "Attribute: %s, Existing value: %s, Requested value: %s" % [attr, existing_attr_val, requested_val]
       Puppet.debug(message)
       if existing_attr_val && requested_val
         seq_diff = requested_val.delete(' ').split(',').zip(existing_attr_val.delete(' ').split(',')).select{|new_val, exist_val| new_val != exist_val}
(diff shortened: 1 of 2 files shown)


gavin-scott:
@logankimmel looks like this will need to be rebased.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -1036,5 +1036,36 @@ def fc630_raid_configuration
     }
   end
 
+  def wsman
+    require "asm/wsman"
+    @wsman ||= begin
+                 logger = Logger.new(STDOUT)
+                 endpoint = {:host => @ip, :user => @username, :password => @password}
+                 ASM::WsMan.new(endpoint, :logger => logger)
+
```
gavin-scott: No blank line.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -1036,5 +1036,36 @@ def fc630_raid_configuration
     }
   end
 
+  def wsman
+    require "asm/wsman"
+    @wsman ||= begin
+                 logger = Logger.new(STDOUT)
+                 endpoint = {:host => @ip, :user => @username, :password => @password}
+                 ASM::WsMan.new(endpoint, :logger => logger)
```
gavin-scott: Generally we pass `Puppet` for the logger from puppet modules. `ASM::WsMan` will handle logging to that, i.e. `Puppet.info`, etc.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -1036,5 +1036,36 @@ def fc630_raid_configuration
...
+  #
+  # This will make a wsman call to the current server to get the
+  # the attribute at this point in time
+  #
+  # @param [:biosbootseq, :hddseq] attribute attribute we are looking for
+  # @return [String] the attribute value
+  def find_current_boot_attribute(attribute)
+    boot_source_settings = wsman.boot_source_settings
+    if attribute == :hddseq
+      sorted_hdd = boot_source_settings.select{ |b| b[:boot_source_type] == "BCV" && b[:current_enabled_status] == "1"
```
gavin-scott: For multi-line please do `do` / `end` instead of `{}`

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -1036,5 +1036,36 @@ def fc630_raid_configuration
     }
   end
 
+  def wsman
+    require "asm/wsman"
+    @wsman ||= begin
+                 logger = Logger.new(STDOUT)
```
gavin-scott: I know this is the way IntelliJ formats, but I think rubocop would prefer two spaces in from the `@wsman` line.

lib/puppet/provider/importtemplatexml.rb
```diff
@@ -318,9 +318,9 @@ def munge_config_xml
     end
 
     %w(BiosBootSeq HddSeq).each do |attr|
-      existing_attr_val = find_bios_attribute(xml_base, attr)
+      existing_attr_val = find_current_boot_attribute(attr.downcase.to_sym)
       requested_val = @changes['partial']['BIOS.Setup.1-1'][attr]
-      message = "Attribute: %s, Existing value: %s, Requested value: %s" % [attr, existing_attr_val, requested_val] 
+      message = "Attribute: %s, Existing value: %s, Requested value: %s" % [attr, existing_attr_val, requested_val]
```
gavin-scott: It seems like there have been many, many attempts to fix the problem of setting HddSeq or BiosBootSeq to an existing value. In addition to this section there is another section here with similar logic:

https://github.com/logankimmel/dell-idrac/blob/ecf24e1fdb8a04e7a09eb3e1f5776e673b3674c6/lib/puppet/provider/importtemplatexml.rb#L390-L411

It's not clear to me whether your changes here are good enough or if the above linked code needs to be tweaked as well....

logankimmel:
@gavin-scott This has been updated.

lib/puppet/provider/importtemplatexml.rb
logankimmel: I don't think my code should be invoked in that other location

logankimmel:
Can somebody review this so I don't have to keep rebasing it ?

PR #261, opened 2016-11-01
ASM-8257 iDRAC configuation failing for VSAN and SD card deployments

sushilrai:
With recent changes for converting the non-RAID disk to RAID mode, VSAN and SD card deployment got broken as there is no raid_confiuraiton parameter in the asm::idrac resource


aimers1975:
@gavin-scott or @ripienaar  Can you take a look, this is blocking VSAN testing.

PR #263, opened 2016-11-30
ASM-8438 ImportSystemConfig runs forever

logankimmel:
Even though we were doing a lookup on the current HDDSeq and
BIOSBootSeq (not using the export xml)  We were only generating
the config xml one time.  Not we will regenerate the config xml
on every importsystemconfig run and rotate the old config (should
only be max of 3)

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -11,6 +11,7 @@ include REXML
 class Puppet::Provider::Importtemplatexml <  Puppet::Provider
 
   attr_accessor :embedded_sata_change
+  attr_accessor :attempt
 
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -373,6 +374,11 @@ def munge_config_xml
     remove_invalid_settings(xml_base)
     # Disable SD card and RAID controller for boot from SAN
 
+    # Rotate the old xml files
+    unless attempt == 0
```
kaushalgala: I don't mind using unless, but it might be easier to read  `if attempts != 0` or `if attempts > 0`

gavin-scott:
Merging as is, @logankimmel preferred it as is and I don't have a strong preference...

PR #264, opened 2016-12-01
ASM-8442 IDRAC able to try importsystemconfiguration for 4 times

prem548:
When IDRAC trying to apply import system configuration settings sometimes  configurations will be not be in sync so IDRAC will try to apply settings for couple of times. 

The below change will try to keep it in sync by trying applying configuration settings for four times,

### DIFF
--- a/lib/puppet/provider/importsystemconfiguration/default.rb
+++ b/lib/puppet/provider/importsystemconfiguration/default.rb
@@ -68,10 +68,10 @@ Puppet::Type.type(:importsystemconfiguration).provide(
     rescue Puppet::Idrac::ConfigError => e
       attempts += 1
       case attempts
-      when 1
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/importsystemconfiguration/default.rb
```diff
@@ -68,10 +68,10 @@ def import_template_job(obj)
     rescue Puppet::Idrac::ConfigError => e
       attempts += 1
       case attempts
-      when 1
-        Puppet.info("First import failed.  Retrying import....")
+      when 1, 2
+        Puppet.info(" %s importsystem failed.  Retrying importing...." % attempts)
```
gavin-scott: Maybe do `'Attempt #%d"`

kaushalgala: +1

PR #268, opened 2016-12-23
ASM-8548 Wait for failed import jobs to complete

gavin-scott:
Existing ImportSystemConfiguration job polling would exit early
(before the job says that it is 100% complete) in the case of
failure. Because of that and the fact that Importtemplatexml did not
wait for LC ready prior to cacluating the configuration to apply it
was possible that a previous ImportSystemConfiguration job was in
progress when a new import was begun. Therefore code that was
checking iDrac inventory might be getting values that would change
as the result of a previously running job. This could lead to the
dreaded "ImportSystemConfiguration job runs forever" bug which
happens in iDrac prior to 2.40.40.40 when a BiosBootSeq or HddSeq
is set to a pre-existing value (i.e. not changed).

This commit attempts to fix this by:

- waiting for LC ready prior to calculating the configuration to
  apply

- waiting for failed jobs to report 100% complete

- waiting for LC ready after all failed import jobs

### DIFF
--- a/lib/puppet/provider/checkjdstatus.rb
+++ b/lib/puppet/provider/checkjdstatus.rb
@@ -7,22 +7,21 @@ class Puppet::Provider::Checkjdstatus < Puppet::Provider
     @instanceid = instanceid
   end
 
-  def checkjdstatus
+  def wsman
     require 'asm/wsman'
-    endpoint = {:host => @ip, :user => @username, :password => @password}
-    schema = "http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/root/dcim/DCIM_LifecycleJob?InstanceID=#{@instanceid}"
-    job_status, job_message, message_id = ASM::WsMan.invoke(endpoint, 'get', schema,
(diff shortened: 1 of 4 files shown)


gavin-scott:
DO NOT MERGE. Need to do end-to-end testing and see if CCB wants to merge this into 8.3.1

gavin-scott:
End to end testing has passed. Was not able to actually reproduce the original failure scenario but had a successful server deployment.

@sushilrai  or @ripienaar can you review but not merge? This needs to go through CCB before being merged into 8.3.1

ripienaar:
Seems good to me, but this isnt really things I know about, @sushilrai maybe

sushilrai:
+1

gavin-scott:
CCB decided to defer for 8.3.1. Re-targeted this to the master branch.

gavin-scott:
Oops, this had the wrong jira on it. Update the title to the correct one but the commit will still be wrong :-(

PR #269, opened 2017-01-03
ASM-8438, ASM-8545 Wait for failed import jobs to complete

gavin-scott:
Existing ImportSystemConfiguration job polling would exit early
(before the job says that it is 100% complete) in the case of
failure. Because of that and the fact that Importtemplatexml did not
wait for LC ready prior to cacluating the configuration to apply it
was possible that a previous ImportSystemConfiguration job was in
progress when a new import was begun. Therefore code that was
checking iDrac inventory might be getting values that would change
as the result of a previously running job. This could lead to the
dreaded "ImportSystemConfiguration job runs forever" bug which
happens in iDrac prior to 2.40.40.40 when a BiosBootSeq or HddSeq
is set to a pre-existing value (i.e. not changed).

This commit attempts to fix this by:

- waiting for LC ready prior to calculating the configuration to
  apply

- waiting for failed jobs to report 100% complete

- waiting for LC ready after all failed import jobs

### DIFF
--- a/lib/puppet/provider/checkjdstatus.rb
+++ b/lib/puppet/provider/checkjdstatus.rb
@@ -7,22 +7,21 @@ class Puppet::Provider::Checkjdstatus < Puppet::Provider
     @instanceid = instanceid
   end
 
-  def checkjdstatus
+  def wsman
     require 'asm/wsman'
-    endpoint = {:host => @ip, :user => @username, :password => @password}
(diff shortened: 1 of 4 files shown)


gavin-scott:
This is an 8.3.1 back-port of https://github.com/dell-asm/dell-idrac/pull/268

It was also observed that without this change we sometimes see intermittent UTF-8 parsing errors when calling `wsman.boot_source_settings`. That appears to be because we are calling it while the iDrac is in the middle of changing the boot settings due to an ImportSystemConfiguration job. (because before this change we did not always wait for the job to complete).

PR #271, opened 2017-02-04
Revert "ASM-8606 Replaced nokogiri 1.5.10 with 1.6.8 version for puppet 3.8"

gavin-scott:
Reverts dell-asm/dell-idrac#270


gavin-scott:
Tests won't pass until dell-asm/dell-asm-util#85 is merged due to the overly strict gem requirements in dell-asm-util.

PR #272, opened 2017-04-10
ASM-6823 Ability to place Raid Mode into pass-through with Hardware configuration only

sushilrai:
For VSAN deployment of SD card, HBA mode configuration was getting skipped as empty RAID configuration is passed from asm-deployer

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,8 +17,8 @@ end
 
 group :development, :test do
   gem 'rake'
-  gem 'rspec'
-  gem 'puppetlabs_spec_helper'
+  gem 'rspec', '~>3.4.0', :require => false
+  gem 'puppetlabs_spec_helper', '0.4.1', :require => false
   gem 'json_pure', '2.0.1'
   if puppetversion = ENV['PUPPET_GEM_VERSION']
     gem 'puppet', puppetversion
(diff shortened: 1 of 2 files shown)


kaushalgala:
@gavin-scott I reviewed this PR and looks fine,

The TravisCI failure seems unrelated: Gem::InstallError: parallel_tests requires Ruby version >= 2.0.0

aimers1975:
@gavin-scott Can I merge this?

gavin-scott:
@sushilrai we talked the other day with @aimers1975 and it was clear we didn't understand what the root problem was. I.e. VSAN deployments also had no raid_configuration parameter but were able to successfully set HBA mode. Did you figure out what the difference was?

gavin-scott:
Travis CI failures are probably due to a gem dependency update. It would be good to pin whatever it is to a compatible version.

sushilrai:
@gavin-scott @kaushalgala tests are successful now. Please review

gavin-scott:
Thanks @sushilrai I'm still unclear as to why this was failing to begin with. Do you think the problem was that VSAN with SD in general wasn't setting HBA mode?

sushilrai:
Yes, that is understanding. For SD Card based deployment RAID_CONFIGURATION is empty hash passed to the puppet module.

PR #276, opened 2018-02-12
LUD-94 Fix XML parsing issue introduced by version upgrades

gavin-scott:
ActiveSupport's `Hash.from_xml` started failing in
`find_target_bios_setting` after recent ruby and gem upgrades with the
error `String can't be coerced into Integer`. Changed this method to
use `ASM::WsMan#bios_enumerations` which already has parsing for
this. Note that the return value of `find_target_bios_setting` has
changed slightly but all callers only use the return value to check
whether it is `nil` or not.


mdcondon:
👍

PR #279, opened 2018-03-07
LUD-222 Support BOSS boot device which includes scale io RAID config.

aimers1975:
Updated backend to support RAID configuration passed into hardware configuration for local flash storage servers.


aimers1975:
@sushilrai Could you merge these changes while Gavin is out?  He's already reviewed.

PR #281, opened 2018-03-21
LUD-407 Fix bug with null exception in non-ScaleIO case

aimers1975:
Fix bug with null exception in non-ScaleIO case for local flash storage. Need to create the full RAID config when the scale IO raid configuration is not present.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -590,10 +590,14 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
 
     @raid_configuration ||=
         begin
-          unprocessed = @resource[:raid_configuration]
+          unprocessed = @resource[:raid_configuration] || {}
+          %w(virtualDisks externalVirtualDisks externalSsdHotSpares externalHddHotSpares ssdHotSpares hddHotSpares).each do |key|
+            unprocessed[key] ||= []
+          end
           #For Local Flash boot device with BOSS we need to add RAID 1
           #configuration to our existing RAID configuration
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -594,7 +594,18 @@ def raid_configuration
           #For Local Flash boot device with BOSS we need to add RAID 1
           #configuration to our existing RAID configuration
           if boss_controller && @boot_device =~ /LOCAL_FLASH_STORAGE/i
-            unprocessed["virtualDisks"].push(boss_virtual_disk)
+            # If there is an existing RAID configuration, we just want to add boss virtual disk
+            # if not we need to create the full RAID config for BOSS.
+            if !unprocessed["virtualDisks"].nil?
```
gavin-scott: There's already a lot of ugly code in this method around handling possibly uninitialized `unprocessed` data. Instead of doing this how bad just adding some initialization code around line 593. Something like:

```
unprocessed = @resource[:raid_configuration]
unprocessed ||= {}
%w(virtualDisks externalVirtualDisks externalSsdHotSpares externalHddHotSpares ssdHotSpares hddHotSpares).each do |key|
  unprocessed|key| ||= []
end
```

Then you can leave your code how it was originally, just `unprocessed["virtualDisks"].push(boss_virtual_disk)`. Also if you're up for it you can remove the rest of the `unprocessed.nil?` checks later in this method since it will never be nil.

aimers1975: @gavin-scott Sounds good, let me make those fixes!

aimers1975:
@gavin-scott Thanks for the feedback, its updated...let me know if you see any other issues.

PR #282, opened 2018-03-23
LUD-534 Deployments are failing for 13g nonRAID configurations

aimers1975:
nonRAID configuration was failing as logic for determining if non-raid disk was requested was wrong, and an extra nonRAID config was getting created.  Also, we remove the nonRAID config when the
controller does not support it, as in the case of HBA330 Mini.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -418,7 +418,6 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
     unless attempt == 0
       rotate_config_xml_file
     end
-
     File.open(@config_xml_path, 'w+') do |file|
       if embsata_in_sync?
         file.write(xml_base.to_xml(:indent => 2))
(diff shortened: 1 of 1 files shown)


gavin-scott:
@aimers1975 this is breaking 13G RAID configurations for me. It appears to be because it's now trying to do some configuration on a RAID controller that wasn't mentioned in the deployment.json:

GetConfigResults error message:

    <ConfigResults>
      <JobID>JID_219339745183</JobID>
      <JobName>Import Configuration</JobName>
      <JobDisplayName>Configure: Import system configuration XML file</JobDisplayName>
      <FQDD>RAID.Embedded.2-1</FQDD>
      <Messages>
        <MessageID>SYS030</MessageID>
        <Message>Controller not found</Message>
      </Messages>
    </ConfigResults>

Offending bit of generated system configuration:

    <Component FQDD="RAID.Embedded.2-1">
      <Attribute Name="RAIDresetConfig">True</Attribute>
      <Attribute Name="RAIDforeignConfig">Clear</Attribute>
      <Component FQDD="Disk.Direct.5-9:RAID.Embedded.2-1">
        <Attribute Name="RAIDPDState">Ready</Attribute>
      </Component>
    </Component>

I won't revert this for now on the theory that maybe @sushilrai 's non-raid deployments will succeed with these changes.

I'll attach more details to the jira ticket.

PR #285, opened 2018-04-02
LUD-534 Fixed bug where removing nonraid disks

aimers1975:
Fixed bug where removing nonraid disks was not happening for nonRAID controllers that don't
support it.


aimers1975:
@gavin-scott Here's the fix for that nonraid disk issue.

PR #286, opened 2018-05-23
LUD-939 User#1 is anonymous and cannot be changed. Lookup other 15 us…

taqeem:
…ers for the 'root' user availability.

         (1)If 'root' user is available at any of the positions from #2 to #16, let it be.
         (2)If 'root' user is NOT available at any of the positions from #2 to #16, set the 'root' user at #2 position as it is reserved for 'root' user

### DIFF
--- a/lib/puppet/provider/server_onboard/default.rb
+++ b/lib/puppet/provider/server_onboard/default.rb
@@ -37,9 +37,24 @@ Puppet::Type.type(:server_onboard).provide(:default, :parent => Puppet::Provider
       raise "username or password not specified"
     end
 
-    Puppet.debug("Setting credentials for %s" % resource[:name])
-    racadm_set("idrac", "users", "username", credential["username"], "2", :raise_on_err => true)
-    racadm_set("idrac", "users", "password", credential["password"], "2", :raise_on_err => true)
+    root_slot = nil
+    empty_slot = nil
+
+    (2..16).each do |i|
+      user_name = racadm_get('idrac', 'users', 'username', i )
+      root_slot = i if user_name == "root"
+      break if root_slot
+      empty_slot = i unless user_name
+    end
+
+    if root_slot
+      racadm_set("idrac", "users", "password", credential["password"], root_slot, :raise_on_err => true)
+    elsif empty_slot
+      racadm_set("idrac", "users", "username", credential["username"], empty_slot, :raise_on_err => true)
+      racadm_set("idrac", "users", "password", credential["password"], empty_slot, :raise_on_err => true)
+    else
+      raise("No root user found and no empty slots available") if !root_slot && !empty_slot
+    end
   end
 
   # Set network configuration
--- a/spec/unit/puppet/provider/server_onboard/server_onboard_spec.rb
+++ b/spec/unit/puppet/provider/server_onboard/server_onboard_spec.rb
@@ -19,6 +19,8 @@ describe Puppet::Type.type(:server_onboard).provider(:default) do
     end
 
     it "should invoke racadm_set commands for specified credentials" do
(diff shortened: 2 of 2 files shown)


gavin-scott:
@sushilrai can you review?

lib/puppet/provider/server_onboard/default.rb
```diff
@@ -36,10 +36,23 @@ def setup_credential
...
+    # set the 'root' user at #2 position as it is reserved for 'root' user
+    is_root_user_present = false
+    (2..16).each do |i|
+      user_name = racadm_get('idrac', 'users', 'username', i )
+      if user_name == credential["username"]
+        is_root_user_present = true
+        break
+      end
+    end
+    unless is_root_user_present
```
gavin-scott: If the root user _is_ present don't you need to set the password?

taqeem: Fixed.

taqeem:
@sushilrai Further to our discussion regarding running tests locally, please do help us in fixing these test fails.

@gavin-scott @naveenhirebeeldell FYI..

taqeem:
@gavin-scott @sushilrai As discussed the unit tests failure is not due to changes made as part of this defect fix. Kindly approve the PR.

 @naveenhirebeeldell FYI

lib/puppet/provider/server_onboard/default.rb
```diff
@@ -36,10 +36,24 @@ def setup_credential
...
+      if user_name == credential["username"]
+        racadm_set("idrac", "users", "password", credential["password"], i, :raise_on_err => true)
+        is_root_user_present = true
+        break
+      end
+    end
+    unless is_root_user_present
+      Puppet.debug("Setting credentials for %s" % resource[:name])
+      racadm_set("idrac", "users", "username", credential["username"], "2", :raise_on_err => true)
+      racadm_set("idrac", "users", "password", credential["password"], "2", :raise_on_err => true)
```
gavin-scott: It seems like we haven't ensured that no other user is already in position 2, the code above just guarantees that user 2 is not `root`. It seems like the code should be something like:

    root_slot = nil
    empty_slot = nil

    (2..16).each do |i|
      user_name = racadm_get('idrac', 'users', 'username', i )

      root_slot = i if user_name == "root"
      break if root_slot

      empty_slot = i unless user_name # double-check: user_name might be an empty string instead of nil for empty slots
    end

    if root_slot
      racadm_set("idrac", "users", "password", credential["password"], root_slot, :raise_on_err => true)
    elsif empty_slot
      racadm_set("idrac", "users", "username", credential["username"], empty_slot, :raise_on_err => true)
      racadm_set("idrac", "users", "password", credential["password"], empty_slot, :raise_on_err => true)
    else
      raise("No root user found and no empty slots available") if !root_slot && !empty_slot
    end

taqeem: @gavin-scott That part of the code that set the root user creds at #2 position was EXISTING even code before we took this defect. And I agree that it replaces the user at #2 with root user. We, as part of this defect fix, are only ensuring that happens conditionally if there is NO root user at any other position. 

The problem you are talking about is very valid but completely different from the scope of this defect.

Considering this I request you to kindly merge this PR and create a new defect for the problem you mentioned.  

@naveenhirebeeldell fyi

gavin-scott: @taqeem the goal here is not to close tickets but to improve the software. Since you're changing this code anyway and we both agree that its still not correct we need to go ahead and fix it, especially since it's not a heavy lift.

taqeem:
@gavin-scott Updated changes as discussed on this the other day. Please merge this PR. Thank you!

PR #288, opened 2018-05-29
Lud 939 temp

taqeem:
Temporary PR. Will be closed


sushilrai:
@taqeem  are you done with these changes? seems you have found the correct compatible version of mocha

taqeem:
@sushilrai Yes, I used mocha 1.4.0 version as suggested by Gavin and it seems to be working. I will check with Shoaib if we have to commit with same JIRA id or a new defect will be created.

PR #290, opened 2018-06-06
LUD-1088 Fixed Issue with Raid config override issue on retry

arunrordell:
Previously export_job ignores virtual_disk information and creates raid
config thinking that it was'nt present even though it already exists.

This PR includes virtual disk info in export_config file


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -772,7 +772,7 @@ def raid_in_sync?(xml_base, log=false)
       raid_configuration.keys.each do |raid_fqdd|
         raid_fqdd_xpath = "//Component[@FQDD='#{raid_fqdd}']"
         controller_xml = xml_base.xpath(raid_fqdd_xpath)
-        existing_virtual_disks = controller_xml.xpath("Component[starts-with(@FQDD, 'Disk.')]")
+        existing_virtual_disks = controller_xml.xpath("Component[starts-with(@FQDD, 'Disk.Virtual')]")
```
gavin-scott: Was there an issue with just checking `Disk.`?

PR #291, opened 2018-06-07
LUD-937 Retry checking_lc_Status in case of wsman Timeout error

arunrordell:


lib/puppet/provider/checklcstatus.rb
```diff
@@ -11,7 +11,16 @@ def initialize (ip, username, password)
   end
 
   def checklcstatus
-    executelccmd
+    execution_count = 0
+    begin
+      executelccmd
+    rescue ASM::WsMan::Error => e
+      if execution_count < 2
+        sleep(10)
```
gavin-scott: If we're getting wsman timed out errors, things are probably running pretty slow on the iDrac, so I think this may not be waiting long enough. You could try using the `ASM::Util.block_and_retry_until_ready` helper for this:

    # This should try to run executelccmd for up to 5 minutes, max wait time of 1 minute and retry on ASM::WsMan::Error
    ASM::Util.block_and_retry_until_ready(300, ASM::WsMan::Error, 60, Puppet, &:executelccmd)

arunrordell:
@gavin-scott  updated this one please review

PR #292, opened 2018-06-27
LUD-1531 fix iDRAC onboarding with static IP address

danielseankim:
It looks like we pass array of network objects, instead of a
single network object. Generally, this array contains only one
network object. This change addresses the handling of the input
parameter type, when a server is being onboarded.

### DIFF
--- a/lib/puppet/provider/server_onboard/default.rb
+++ b/lib/puppet/provider/server_onboard/default.rb
@@ -70,11 +70,10 @@ Puppet::Type.type(:server_onboard).provide(:default, :parent => Puppet::Provider
 
     if network_type.to_s == "static"
       network_obj = resource[:networks]
-      if network_obj["staticNetworkConfiguration"].nil? || network_obj["staticNetworkConfiguration"].empty?
-        raise "network configuration params are not static"
+      Array(network_obj).each do |net|
+        next if net["staticNetworkConfiguration"].nil? || net["staticNetworkConfiguration"].empty?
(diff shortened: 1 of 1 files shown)


danielseankim:
@gavin-scott can you review/merge please?

lib/puppet/provider/server_onboard/default.rb
```diff
@@ -70,11 +70,10 @@ def setup_network
 
     if network_type.to_s == "static"
       network_obj = resource[:networks]
-      if network_obj["staticNetworkConfiguration"].nil? || network_obj["staticNetworkConfiguration"].empty?
-        raise "network configuration params are not static"
+      network_obj.each do |net|
```
gavin-scott: If there is any concern that this isn't always an array you could do `Array(network_obj).each do |net|` which would work either way.

danielseankim:
@gavin-scott I made the change requested here.

lib/puppet/provider/server_onboard/default.rb
```diff
@@ -70,11 +70,10 @@ def setup_network
 
     if network_type.to_s == "static"
       network_obj = resource[:networks]
-      if network_obj["staticNetworkConfiguration"].nil? || network_obj["staticNetworkConfiguration"].empty?
-        raise "network configuration params are not static"
```
gavin-scott: @danielseankim why did you remove this `raise`? Seems like this change breaks the tests, but looks like Travis CI didn't run on this unfortunately.

PR #294, opened 2018-08-01
VXFM-1531 Fix unit test failures from previous change

gavin-scott:


gavin-scott:
Fix for unit test failures introduced by https://github.com/dell-asm/dell-idrac/pull/292

PR #295, opened 2018-08-01
VXFM-2163 Fix server_onboard for default password warning

gavin-scott:
After an iDrac factory reset it will report a warning banner when
logging in via ssh that says "Warning: It is recommended not to use
the default user name (root) and password as it is a security
risk.". This broke the `racadm_get` parsing and thus the
`server_onboard` idrac user password setting which uses racadm.


gavin-scott:
Tests will fail until this one is merged: https://github.com/dell-asm/dell-idrac/pull/294

PR #298, opened 2019-02-05
VXFM-4166 Skip NVMe drives without controllers

gavin-scott:
NVMe drives may be connected directly to the PCIe bus without a
controller. However, they are passed in the `raid_configuration` with
a "fake" controller value of "Enclosure.Internal.0-1" in the
`virtualDisks` list:

    raid_configuration:
      virtualDisks:
      - raidLevel: nonraid
        physicalDisks:
        - Disk.Bay.21:Enclosure.Internal.0-1
        - Disk.Bay.22:Enclosure.Internal.0-1
        - Disk.Bay.23:Enclosure.Internal.0-1
        controller: Enclosure.Internal.0-1
        configuration:
          id: 94e9ea6e-80a6-4bc1-be69-541334a30e84
          raidlevel: nonraid
          comparator: minimum
          numberofdisks: 3
          disktype: requirenvme
          controllerFqdd: Enclosure.Internal.0-1
        mediaType: NVMe

In this case they need to be skipped in the same way as
HBA330-attached disks -- they require no configuration.

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -585,19 +585,16 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
       @nonraid_to_raid = true
       @resource[:raid_configuration]["virtualDisks"] << non_raid_disks
     end
+
     # Check that any non-raid virtual disks are being requested on a controller that supports configuration.
     # Currently only the HBA330 Mini does not support configuration and should already be in non-raid
     # mode.
     unless @resource[:raid_configuration].nil? || @resource[:raid_configuration]["virtualDisks"].nil?
-      @resource[:raid_configuration]["virtualDisks"].each do |this_virtual_disk|
-        if this_virtual_disk["raidLevel"] == "nonraid"
-          if !controller_supports_non_raid?(this_virtual_disk["controller"])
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -1067,7 +1064,7 @@ def is_embedded_raid?
   # @return Boolean
   def controller_supports_non_raid?(non_raid_fqdd)
     non_raid_disk_controller = disk_controllers.find { |c| c[:fqdd].include?(non_raid_fqdd) }
-    !(non_raid_disk_controller[:product_name] =~ /Dell HBA330/i)
+    !(non_raid_disk_controller.nil? || non_raid_disk_controller[:product_name] =~ /Dell HBA330/i)
```
saxens7: We will have to handle NVMe 740 nodes that are connected to controller. We will have to add :productname exception for 'PCIeExtender' as well.

gavin-scott: Does the extender come back from the `DCIM_ControllerView` enumeration / `disk_controllers` method? If so then yes an exception needs to be added here. If you send the product name I can update this PR for you.

saxens7: I thought we were getting controller info for 740's but thats not the case. Current check will work for 740 aswell

gavin-scott:
Merging based on @saxens7 review.

PR #304, opened 2019-04-26
VXFM-5951 Need to sanitize NVDIMMs during new deployment

sushilrai:
In case of NVDIMM on server, then we need to set attribute on BIOS section to enable sanitize NVDIMMs


sushilrai:
Dependent PR https://github.com/dell-asm/dell-asm-util/pull/125

sushilrai:
Message `The erase operation on the non-volatile DIMM with serial number 181D8D4D in slot B5 is successfully completed.` is listed in the LC logs when `import system configuration` is performed.

sushilrai:
Yes, considering the behavior for initialization process Import operation will run everytime for NVDIMM enabled server

PR #312, opened 2019-08-18
VXFM-6731 VXFM-6732 modify discovered Idrac to set snmp options

credzba:
Modified the idrac discovery to have the option to set the snmp alerts and set idrac to send alerts to the flex manager.


credzba:
going to put changes in non-open source path

PR #314, opened 2019-10-09
[DNM] VXFM-6427 handle temporary fault errors

danielseankim:


danielseankim:
With the changes in dell-asm-util, we need to handle the new type of errors for temporary errors. Test is in progress, and the DNM tag will be removed once it completes successfully.

lib/puppet/provider/idrac_fw_installfromuri/wsman.rb
```diff
@@ -78,7 +78,7 @@ def update(firmware_list)
         sleep 30
         begin
           statuses[job_id][:status] = checkjobstatus job_id
-        rescue ASM::WsMan::Error, RuntimeError => e
+        rescue ASM::WsMan::Error, ASM::WsMan::FaultError, RuntimeError => e
```
gavin-scott: It seems like maybe `FaultError` and `ResponseError` should extend `Error` since they are types of wsman error. That would fix this as well and maybe fix other places.

danielseankim: I opened https://github.com/dell-asm/dell-asm-util/pull/150. Feel free close this one once that PR is merged.

gavin-scott:
Closed in favor of https://github.com/dell-asm/dell-asm-util/pull/150

PR #315, opened 2019-11-08
VXFM-8464 Make changes to support raid boot for nodes in a deployment

kannawin:
This line was creating problems in the idrac config file causing deployments to fail because it was setting a bad value for hddseq as thats already set with pxe_boot options

Requires Merging from these PR's at same time:
https://github.com/dell-asm/dell-asm-util/pull/158
https://github.com/dell-asm/asm-deployer/pull/3145
https://github.com/dell-asm/asm-core/pull/5433

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -705,7 +705,7 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
             changes['whole'][raids.first] = { 'CurrentControllerMode' => "RAID" }
           end
         end
-        changes['partial'] = {'BIOS.Setup.1-1'=> {'HddSeq' => raid_configuration.keys.first}} if @boot_device =~ /HD/i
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -705,7 +705,7 @@ def get_raid_config_changes(target_current_xml, raid_reset=false)
             changes['whole'][raids.first] = { 'CurrentControllerMode' => "RAID" }
           end
         end
-        changes['partial'] = {'BIOS.Setup.1-1'=> {'HddSeq' => raid_configuration.keys.first}} if @boot_device =~ /HD/i
```
gavin-scott: Why was this removed? How do we ensure that the correct device is first in the `HddSeq` without this?

kannawin: afaik with UEFI boot, and with the "HD" value now used the enable_pxe sets boot order. I can go back and try a value for this that works. It was causing every RAID deployment to error with invalid characters on this line of the .yaml config file for the idrac.

PR #316, opened 2019-11-11
VXFM-9100 Automated iDrac reset was failing due to Net::SSH changes

gavin-scott:
`Net::SSH::Verifiers::Null` was removed. It looks like
`:verify_host_key => :never` is the more correct way to disable host
key verification but for Dacite release sticking with
`:verify_host_key => false` which is already in use in existing code.


gavin-scott:
Marked DNM while end-to-end test in progress.

PR #317, opened 2020-01-28
VXFM-9988 new sata dom returned different device string

credzba:
Some new satadom were returning a different device string that did not contain the words /SATADOM/ and as such were unrecognized

### DIFF
--- a/lib/puppet/provider/importtemplatexml.rb
+++ b/lib/puppet/provider/importtemplatexml.rb
@@ -310,11 +310,9 @@ class Puppet::Provider::Importtemplatexml <  Puppet::Provider
 
   def get_boot_sata_disk
     # If we find a SATADOM device, prefer it, else fallback to getting first SATA disk
-    boot_sources = Puppet::Idrac::Util.boot_source_settings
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/importtemplatexml.rb
```diff
@@ -311,7 +311,7 @@ def is_sd_card?
   def get_boot_sata_disk
     # If we find a SATADOM device, prefer it, else fallback to getting first SATA disk
     boot_sources = Puppet::Idrac::Util.boot_source_settings
-    satadom_instance_id = boot_sources.select {|d| d[:boot_string] =~ /SATADOM/}[0][:instance_id] rescue nil
+    satadom_instance_id = boot_sources.select {|d| d[:boot_string] =~ /SATADOM/ || d[:boot_string] =~ /Embedded SATA Port Disk [A-Z]/}[0][:instance_id] rescue nil
     if satadom_instance_id
       Puppet.debug("Prefering SATADOM disk for booting")
       satadom_instance_id.split("#")[2]
```
gavin-scott: It looks like above 5 lines of code are identical to the `get_satadom` method. Can you clean it up to just call that method instead of duplicating the logic?

credzba: Done

